PR #32, opened 2014-11-11
SI-4520 - reasonable exception instead of OOE in XML parser

terma:
That's fix `OutOfMemoryException` for case when we parse valid XML, however user code request not existent content or not correct content type. Now parser instead of consume all memory just raise exception that request info not present.

Jira: https://issues.scala-lang.org/browse/SI-4339 and https://issues.scala-lang.org/browse/SI-4520


SethTisue:
anyone interested in moving forward on this...?

ashawley:
Re-opened as #209 to see if these tests could be fixed, including some additional tests.

PR #50, opened 2015-04-11
Fix default namespace in PrettyPrinter.formatNodes

chrisloy:
This replicates the fix from #28 and applies it to the ```formatNodes``` method of ```PrettyPrinter``` as well. Existing behaviour is the same bug that @ashawley highlighted before - namely the addition of an erroneous empty ```xmlns=""``` at the top level of each node.

### DIFF
--- a/src/main/scala/scala/xml/PrettyPrinter.scala
+++ b/src/main/scala/scala/xml/PrettyPrinter.scala
@@ -203,7 +203,7 @@ class PrettyPrinter(width: Int, step: Int) {
    * @param sb   the stringbuffer to append to
    */
   def format(n: Node, sb: StringBuilder) { // entry point
-    format(n, null, sb)
+    format(n, TopScope, sb)
   }
 
   def format(n: Node, pscope: NamespaceBinding, sb: StringBuilder) { // entry point
--- a/src/main/scala/scala/xml/PrettyPrinter.scala
+++ b/src/main/scala/scala/xml/PrettyPrinter.scala
@@ -253,7 +253,7 @@ class PrettyPrinter(width: Int, step: Int) {
    *  @param nodes  the sequence of nodes to be serialized
    *  @param pscope the namespace to prefix mapping
    */
-  def formatNodes(nodes: Seq[Node], pscope: NamespaceBinding = null): String =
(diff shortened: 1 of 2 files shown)


ashawley:
When I investigated #28 I noticed other instances of null default arguments for scope in `PrettyPrinter` that should have been `TopScope`.  I only avoided fixing them to narrow the impact of my patch.  Four months later, it seems there is very little risk fixing these.  Fixing them seems to be the right thing to do.

chrisloy:
Thanks @ashawley. I think the only other one is [here](https://github.com/chrisloy/scala-xml/blob/master/src/main/scala/scala/xml/PrettyPrinter.scala#L206) ... I'm happy to fold that change in too.

ashawley:
I saw that, too.  I'll let @adriaanm determine if he wants a separate PR for that last null.  Perhaps cleaning up the neighborhood is appreciated in this case.

adriaanm:
Happy to let you decide on this! This does seem like a nice occasion to apply the boy scout rule :-)

chrisloy:
I'll have a look this evening (UK time). If it is indeed as simple as it looks then I'll add that to this PR, if not let's do it separately. Thanks guys.

chrisloy:
Second commit added which also fixes (and tests) the behaviour when passing in a ```StringBuilder```. I believe that's the last place!

adriaanm:
Excellent! Btw, our commit guidelines would recommend a title like "Default to TopScope, not null for name space binding."   (The commit stats already tell you which file was updated, and "fix" is too generic.)

chrisloy:
Thanks @adriaanm, have tweaked the commit messages accordingly; I hope they read better now.

Assuming the build passes, do you reckon everything looks in order?

adriaanm:
Looks great to me -- thanks again!

chrisloy:
Hey @adriaanm - just wanted to check if you were waiting on any further action from me before merging this in? I couldn't find a specific contribution policy for this repo, but previous PRs looked to have been merged in with little formal process... Is there anything else you're expecting me to do?

adriaanm:
Sorry about the delay!

PR #55, opened 2015-05-06
Avoid .format in serialization

AlixBa:
avoid .format since it calls java.util.Currency.getInstance and blocks thread

capture d ecran 2015-05-06 a 15 01 35


retronym:
We use `String.format` pervasively in the Scala compiler and standard library, so I'm hesitant to make a localized workaround just here in the XML module.

I'd suggest instead to upgrade to a more recent build of the JVM that contains the fix for https://bugs.openjdk.java.net/browse/JDK-7145454

AlixBa:
Thank you!
I was not aware of this issue in the JVM, I did not find anything on Google.

PR #59, opened 2015-06-30
Fix #58 (SI-4528) BasicTransformer has no longer exponential complexity 

### LINKED ISSUE #58: BasicTransformer has exponential complexity
Please see 
https://issues.scala-lang.org/browse/SI-3689
https://issues.scala-lang.org/browse/SI-4528

It seems BasicTransformer has exponential complexity on the nesting level of the XML being processed. This is due to this method:
```
def transform(ns: Seq[Node]): Seq[Node] = {
  val (xs1, xs2) = ns span (n => unchanged(n, transform(n)))
   
  if (xs2.isEmpty) ns
  else xs1 ++ transform(xs2.head) ++ transform(xs2.tail)
}
```
Each modified node is transformed twice: once at the span, and again at the if/else. So, for <a><b><c><d/></c></b></a>, with c being modified, node a gets transformed twice, node b gets transformed four times (twice for each time a is transformed), and node c gets transformed eight times.

biswanaths:
Replaced the transform(ns) // tranform sequence with one of the proposed methodhttps://issues.scala-lang.org/browse/SI-4528

Added tests to verify that the transform is reusing all the nodes it should be, also equivalency of the original, modified and the alternate option. 

Please let me know, if there are any issue with the pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,9 @@
 Maintainer wanted!
 ==================
 
-[![Join the chat at https://gitter.im/scala/scala-xml](https://badges.gitter.im/Join%20Chat.svg)](https://gitter.im/scala/scala-xml?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
 Would you like to maintain this project? (Please [get in touch](https://github.com/scala/scala#get-in-touch) with someone from the scala/scala core team!)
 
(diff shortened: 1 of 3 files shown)


adriaanm:
LGTM -- would be good to fix typos in commit message, include the longer description from the PR, and mention the github issue as "Fixes #58", which will cause github to cross-link.

I suggest a title like "Fix #58 (SI-4528) BasicTransformer no longer exponential"

adriaanm:
Thanks, @biswanaths! I'll add you to the scala-xml admins so you can merge go through the other PRs and merge those that are ready.

biswanaths:
@adriaanm , updated the commit message to follow the standards as suggested by you.

adriaanm:
Thanks! I'll let you do the honors :)

PR #67, opened 2015-07-24
opt-in to Travis's newer/faster container-based infrastructure

SethTisue:
this worked fine over in scala-partest, presumably will work
fine here too


SethTisue:
review by @adriaanm or @lrytz

adriaanm:
LGTM!

PR #85, opened 2015-10-20
Fix XMLEventReader does not handle &apos; properly

### LINKED ISSUE #72: XMLEventReader does not handle &apos; properly
(This issue migrated from https://issues.scala-lang.org/browse/SI-7796)

Of the five required predefined entities in XML, XMLEventReader does not handle ', returning an EvComment of " unknown entity apos; " instead of an EvEntityRef.

This test program:

``` scala
import scala.xml.pull._
import scala.io.Source
object reader {
  val src = Source.fromString("<test>&amp;&lt;&gt;&apos;&quot;</test>")
  val er = new XMLEventReader(src)
  def main(args: Array[String]) {
    while (er.hasNext)
      Console.println(er.next)
  }
}
```

outputs:

```
EvElemStart(null,test,,)
EvEntityRef(amp)
EvEntityRef(lt)
EvEntityRef(gt)
EvComment( unknown entity apos; )
EvEntityRef(quot)
EvElemEnd(null,test)
```

Also, apos does not appear in XhtmlEntities.scala (may be unrelated).

Since these five entities are predefined, I would argue that the parser should auto-replace them with their equivalents so the user doesn't have to.

fehmicansaglam:
- Fixes #72

### DIFF
--- a/shared/src/main/scala/scala/xml/Utility.scala
+++ b/shared/src/main/scala/scala/xml/Utility.scala
@@ -93,7 +93,7 @@ object Utility extends AnyRef with parsing.TokenTests {
      * For reasons unclear escape and unescape are a long ways from
      * being logical inverses.
      */
-    val pairs = Map(
+    private val pairs = Map(
       "lt" -> '<',
       "gt" -> '>',
       "amp" -> '&',
--- a/jvm/src/test/scala/scala/xml/pull/XMLEventReaderTest.scala
+++ b/jvm/src/test/scala/scala/xml/pull/XMLEventReaderTest.scala
@@ -168,4 +168,40 @@ class XMLEventReaderTest {
     while(er.hasNext) er.next()
     er.stop()
   }
+
+  @Test
+  def entityRefTest: Unit = {
+    val source = Source.fromString("<text>&quot;&apos;&lt;&gt;&amp;</text>")
+    val er = new XMLEventReader(source)
+
+    assertTrue(er.next match {
+      case EvElemStart(_, "text", _, _) => true
+      case _ => false
+    })
+    assertTrue(er.next match {
+      case EvEntityRef("quot") => true
+      case e => false
+    })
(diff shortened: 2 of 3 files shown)


SethTisue:
@fehmicansaglam I'm sorry this seems to have been ignored. it needs a rebase now. are you interested in doing a rebase? perhaps this comment will attract some reviewer attention.

fehmicansaglam:
Sure, I will.

SethTisue:
are this and #89 duplicates?

fehmicansaglam:
@SethTisue yes, they are duplicates. According to your comment [here](https://github.com/scala/scala-xml/issues/72#issuecomment-213100064) #89 is better. If so, I can close this.

SethTisue:
@ashawley I'm confused by your comment on #85. can you make the call here?

ashawley:
@SethTisue This one is good.  I had assumed fixing the XMLEventReader bug would have broader impact.  I was incorrect.  I eventually cleared up my own confusion in [this comment at #72](https://github.com/scala/scala-xml/issues/72#issuecomment-213585916).  So PR #89 was a false lead, but also an attempt to understand the issues.

shared/src/main/scala/scala/xml/Utility.scala
```diff
@@ -93,7 +93,7 @@ object Utility extends AnyRef with parsing.TokenTests {
      * For reasons unclear escape and unescape are a long ways from
      * being logical inverses.
      */
-    val pairs = Map(
+    private val pairs = Map(
```
ashawley: Maybe annotate it as `@deprecated` first just in case a user happened to be importing it?

ashawley:
Fixed instead by 3b0775b.  I used a version of your unit test in 6824df8.  With any luck, this will be released in 1.0.7 relatively soon.

Thanks for looking in to this and submitting the PR.

Sorry it took so long to get this merged.

Issue #86, opened 2015-10-21
Sbt is not able to build with java 1.7 


biswanaths:
On a system with java 1.7, sbt is not able to figure which scala version to use. The reason are the following line from build.sbt. As you can see there is no condition for java 1.7, and it gives the error "don'know what Scala versions to build on $java". Please see https://github.com/scala/scala-xml/blob/master/build.sbt#L11
 
```
crossScalaVersions         := {
  val java = System.getProperty("java.version")
  if (java.startsWith("1.6."))
    Seq("2.11.7", "2.12.0-M1")
  else if (java.startsWith("1.8."))
    Seq("2.12.0-M2")
  else
    sys.error(s"don't know what Scala versions to build on $java")
}
```

SethTisue:
Note that any change must interact properly with `.travis.yml`; see https://github.com/scala/scala-xml/commit/38fbbbeeff6bc97c4438e3e37041894c93538e46.

I guess it should lump 1.7 in with 1.6? (And, while we're at it, lump 1.9 in with 1.8?)

biswanaths:
@SethTisue , Can you expand " change must interact properly with .travis.yml"  ?

If  we add a condition for java1.7 I don't even see a problem with test we are not even testing for java 1.7.  Also is there a reason, why tests1.7 are removed.

I will go ahead and 1.7 to 1.6 and 1.9 to 1.8 .

SethTisue:
there would only be a problem with `.travis.yml` if someone were contemplating a more drastic change to the `crossScalaVersions` logic.

> is there a reason, why tests1.7 are removed

it just seemed like overkill (slower feedback from Travis, and probably no gain., but I don't have a strong feeling about it one way or the other. you could re-enable it if you see value in it.

biswanaths:
Ok, submitted a pr for the same. 

1.9 already. I hardly know anything about 1.8.

PR #87, opened 2015-10-23
Enabling builds for java 1.7 and 1.9

biswanaths:
Please see Issue https://github.com/scala/scala-xml/issues/86

The build file errors if you are trying to build on system with java 1.7. As suggest by @SethTisue , lumped 1.7 with 1.6 and future proofing 1.9 by lumping it together with 1.8.


SethTisue:
LGTM

PR #104, opened 2016-06-12
updated scala version and mima version

biswanaths:


ashawley:
See #98 for the correct fix on this

SethTisue:
@biswanaths should this be closed, since #98 was merged instead?

biswanaths:
Closing this as we have merged #98.

PR #109, opened 2016-09-28
Fix #49: Add Scala.js support

### LINKED ISSUE #49: Can a scala-js version of this be published?
I am not sure if this is feasible, but a scala-js version of this library would help me a lot, since I have a ton of existing code which I could start using on the client-side immediately.

Thanks,
~ hrj

OlivierBlanvillain:
This PR adds Scala.js support.

The changes are pretty straightforward. After updating the bulid, a52ca9c moves files around to cope with the Scala.js `jvm/js/shared` folder structure. Almost everything goes to `/shared`, except for the few test cases that lead to linking errors on Scala.js. Then fc86cb2 copies the problematic tests to `/js` and manually comments the parts involving JVM specific stuff such as accesses to the file system.

I think the prime use case for xml literals in Scala.js is to generate equivalent `org.scalajs.dom.raw.Node`, which does not require any of the JVM specific parts of the library. [Binding.scala](https://github.com/ThoughtWorksInc/Binding.scala) and [scala-js-react](https://github.com/kanterov/scala-js-react) are already doing this by the mean of macros: they get ride of everything coming from `scala.xml._` at compile time. This approach unfortunately leads to long compilation times. Using xml literals natively from Scala.js (probably with a run time transformation to `org.scalajs.dom.raw.Node`) would leads to much better compilation time.


biswanaths:
Thanks @OlivierBlanvillain. I will take a look at the build failure https://travis-ci.org/scala/scala-xml/builds/163345736. And let the community take a look at this too.

biswanaths:
Build seems fine https://travis-ci.org/scala/scala-xml/builds/163353923. There seems intermitted download error.

OlivierBlanvillain:
Yes, I rolled back the version of sbt (I updated it without a good reason anyway), the problem seems to come from repo.typesafe.com (https://github.com/sbt/sbt/issues/2758).

SethTisue:
> I rolled back the version of sbt (I updated it without a good reason anyway)

I'd encourage resubmitting that part once the sbt/sbt#2758 storm has passed

olafurpg:
Wow, this would be a huge enabler for frontend Scala.js development. We are currently using Binding.scala in a medium sized project. Our designer loves using HTML syntax but he hates the abominable compilation times caused by the  macro annotations that eliminate scala.xml._.

One question, it seems `XML.load` is commented out in all the `js/` tests, presumably because it requires JVM specific APIs. Will there be a js alternative to `Xml.load`?

OlivierBlanvillain:
@olafurpg Thanks for your feedback! `XML.load` relies on `SAXParser` which is JVM only. The best option would be to switch to a pure Scala XML parser, this [fastparse](https://github.com/lihaoyi/fastparse/blob/master/scalaparse/shared/src/main/scala/scalaparse/Xml.scala) implementation for example.

kantianethics:
I look forward to this being merged :)

biswanaths:
What would be the right way to test this ? Any ideas ?

sjrd:
> What would be the right way to test this ?

Hum ... `xmlJS/test`?

biswanaths:
@sjrd thank you for pitching in. Really appreciate your view in this.

I have gone through your post http://stackoverflow.com/questions/30143126/how-do-i-port-an-existing-scala-library-to-scalajs/30144349#30144349 and http://www.scala-js.org/doc/project/cross-build.html . 

To be clearer, my question would have been is this sufficient or do we need to have any more tests, to make sure the library works and behaves well when used with scala.js.

sjrd:
It is as sufficient as the tests for the JVM. Everything that should work on both platform should be tested in `shared/src/test/`. They will be tested with `xmlJVM/test` and `xmlJS/test`. I'm not sure what "more" you want. If you want more tests, you can add more tests, but that's just like on the JVM.

biswanaths:
@sjrd thanks for quick reply and regarding tests I got it. I think we should be good enough to go on this then.

build.sbt
```diff
@@ -18,22 +11,26 @@ crossScalaVersions         := {
...
+    List(
+      OsgiKeys.exportPackage := Seq(s"scala.xml.*;version=${version.value}"),
+      libraryDependencies += "junit" % "junit" % "4.11" % "test",
+      libraryDependencies += "com.novocode" % "junit-interface" % "0.10" % "test",
+      libraryDependencies += ("org.scala-lang" % "scala-compiler" % scalaVersion.value % "test").exclude("org.scala-lang.modules", s"scala-xml*"),
+      mimaPreviousVersion := Some("1.0.5")): _*)
+  .jsConfigure(_.enablePlugins(ScalaJSJUnitPlugin))
+
+lazy val xmlJVM = xml.jvm
+lazy val xmlJS = xml.js
```
ashawley: The cleanup of the build.sbt file and modernization for other compilers is great.

xml/js/src/test/scala/scala/xml/XMLTest.scala
```diff
@@ -0,0 +1,877 @@
...
+  //         Elem(null, "title", e, sc, Text("Data on ze web"))),
+  //       Elem(null, "author", e, sc, Text("Peter Buneman")),
+  //       Elem(null, "author", e, sc, Text("Dan Suciu")),
+  //       Elem(null, "title", e, sc, Text("Data on ze web")),
+  //       Elem(null, "book", e, sc,
+  //         Elem(null, "author", e, sc, Text("John Mitchell")),
+  //         Elem(null, "title", e, sc, Text("Foundations of Programming Languages"))),
+  //       Elem(null, "author", e, sc, Text("John Mitchell")),
+  //       Elem(null, "title", e, sc, Text("Foundations of Programming Languages"))))
+  // }
```
ashawley: Again, if no plan to support these tests for the ScalaJS test suite, then just delete instead of comment out?

xml/js/src/test/scala/scala/xml/CompilerErrors.scala
```diff
@@ -0,0 +1,202 @@
+// package scala.xml
```
sjrd: FWIW, I agree.

xml/jvm/src/test/scala/scala/xml/ReuseNodesTest.scala
```diff
@@ -10,19 +10,19 @@ import org.junit.experimental.theories.DataPoints
...
  * This test verify that after the tranform, the resultant xml node
- * uses as many old nodes as possible. 
- * 
- * Three transformers class for case - 
- * One for orginal, one for modified, and one proposed which shows 
+ * uses as many old nodes as possible.
+ *
+ * Three transformers class for case -
+ * One for orginal, one for modified, and one proposed which shows
  * all are equivalent when it comes to reusing as many nodes as possible
```
ashawley: If the only changes to this file, ReuseNodesTests.scala, are changes to whitespace, maybe revert so as to avoid touching the file besides the rename.  We can fix whitespace in another PR

xml/shared/src/main/scala/scala/xml/persistent/CachedFileStorage.scala
```diff
@@ -96,7 +97,7 @@ abstract class CachedFileStorage(private val file1: File) extends Thread {
     val c = fos.getChannel()
 
     // @todo: optimize
-    val storageNode = <nodes>{ nodes.toList }</nodes>
+    val storageNode = <nodes>{nodes.toList}</nodes>
```
ashawley: This file, CachedFileStorage.scala, also only has whitespace changes that should be preserved for now, IMO

build.sbt
ashawley: Will the `publish-signed` task work with the new changes to SBT?  The admin/build.sh file is run by Travis, and when it's for a Git tag, it will copy the admin/publish-settings.sbt file to the root directory.  Looks like it will cause SBT to fail to load:

```
scala-xml/publish-settings.sbt:3: error: not found: value pgpPassphrase
pgpPassphrase := Some(env("PGP_PASSPHRASE").toArray)
^
```

OlivierBlanvillain:
@ashawley thanks for you detailed review! I've forced push 3 updated commits which address all of your comment. Regarding the `publish-signed` did you forgot the `cat admin/gpg.sbt >> project/plugins.sbt` part?

For reference, I managed to `sbt publish-signed` this branche as `"in.nvilla" %%% "scala-xml" % "1.0.6"` using this [build.sbt](https://gist.github.com/OlivierBlanvillain/dc4c0068d7b86acb11edba1be66b6d4c).

sjrd:
> It would be better if it was this, IMO
> src/main/scala/
> src/test/scala/

Whether good or bad, this would against commonly acknowledged practices. Moreover, this would confuse some IDEs.

For the Scala.js thing, it might be better to achieve this:
```
js/src/test/scala/scala/xml/
jvm/src/test/scala/scala/xml/
shared/src/main/scala/scala/xml/
shared/src/test/scala/scala/xml/
```
which can be done with `lazy val xml = crossProject.in(file("."))...`. This removes the unnecessary `xml/` prefix to all the paths.

ashawley:
@OlivierBlanvillain Cool!  Looks great, thanks. Yes, I missed that part of the build.sh script for publish-signed.  

@sjrd I'm familiar with the convention, but wasn't aware the folder structure was a requirement for IDEs. I had assumed they use code indexing and such regardless of folder names to resolve things, anyway.  I'm an SBT and Emacs scala-mode person, so if it's an issue I'd have to defer to others who use such IDEs.

OlivierBlanvillain:
Alright, structure is now as follows:

```
js/src/test/scala/scala/xml/
jvm/src/test/scala/scala/xml/
shared/src/main/scala/scala/xml/
shared/src/test/scala/scala/xml/
```

OlivierBlanvillain:
Rebased. Do you think this could get merged before 1.0.6?

olafurpg:
FWIW, here is a demonstration of what's possible with scala-xml on Scala.js: https://olivierblanvillain.github.io/monadic-html/examples/

The demo uses a fork of scala-xml, it would be nice to depend on the official scala-xml module.

SethTisue:
@biswanaths @ashawley shall we merge?

ashawley:
LGTM.  I appreciate @OlivierBlanvillain's willingness to minimize the file changes, and for entertaining my directory renaming wishes:  Even if in the end, my whims were misguided.

I assume adding Scalacheck in #110 will require a similar pattern of disabling certain tests for ScalaJS?

biswanaths:
@SethTisue , sure enough. 

I think community had enough time to go through this. 

I will go through the changes once again. And pull it in.

ashawley:
Just needs another rebase from master

bbarker:
Any other issues with this? Would really like to see it merged.

bbarker:
@biswanaths @SethTisue @ashawley Nudge, anything I/we can do to help get this in? 

I'd like to see it for scala.js, and once this is in, maybe experiment with using it in scala-native as well.

SethTisue:
(the Scala team doesn't need to be the gatekeeper in this repo, but this an especially juicy PR and there seems to be a consensus on merging it, and certainly plenty of review time passed, so I went ahead in this case)

ashawley:
Travis build error for Java 8 and Scala 2.12:

    unresolved dependency: org.scala-js#scalajs-junit-test-plugin_2.12.1;0.6.12: not found

SethTisue:
huh, I don't know how I missed that, I thought I looked at the Travis results, maybe they were somehow not current. I'll prepare a followup PR right now with a fix.

SethTisue:
thank you, Olivier, for this truly substantial contribution!

ashawley:
So will tests need to be added to both JS and JVM trees going forward?  Was there an option to have SBT run common tests between both Scala compilers?

OlivierBlanvillain:
@ashawley Cross platform tests should go in `shared/src/test`, jvm only tests in `jvm/src/test`. This PR introduced some duplication in tests which is a mistake. Everything in `js/src/test` should have been in `shared/src/test`, and `jvm/src/test` should only contains tests that are not in `shared`.

ashawley:
That's right, `shared`.  Ok, the duplication should be weeded out.

OlivierBlanvillain:
@ashawley I going to open a new PR doing that

PR #122, opened 2016-11-10
Change default encoding in XML.save to UTF-8

Rogach:
Was ISO-8859-1, which resulted in encoding errors at runtime if document
contained non-latin1 characters. Also XML spec states that
documents without xml declaration are expected to contain UTF-8 or UTF-16,
 so writing in ISO-8859-1 without xml declaration (which was the default)
can easily break compliant parsers.

### DIFF
--- a/jvm/src/test/scala/scala/xml/XMLTest.scala
+++ b/jvm/src/test/scala/scala/xml/XMLTest.scala
@@ -300,6 +300,29 @@ class XMLTestJVM {
     </wsdl:definitions>""", wsdlTemplate4("service4", () => "target4") toString)
   }
 
+  // Issue found with ISO-8859-1 in #121 that was fixed with UTF-8 default
+  @UnitTest
+  def writeReadNoDeclarationDefaultEncoding: Unit = {
+    val chars = ((32 to 126) ++ (160 to 255)).map(_.toChar)
+    val xml = <x>{ chars.mkString }</x>
+
+    // com.sun.org.apache.xerces.internal.impl.io.MalformedByteSequenceException:
+    // Invalid byte 1 of 1-byte UTF-8 sequence.
+    // scala.xml.XML.save("foo.xml", xml)
+    // scala.xml.XML.loadFile("foo.xml").toString)
+
+    val outputStream = new java.io.ByteArrayOutputStream
+    val streamWriter = new java.io.OutputStreamWriter(outputStream, XML.encoding)
+
+    XML.write(streamWriter, xml, XML.encoding, false, null)
+    streamWriter.flush
+
+    val inputStream = new java.io.ByteArrayInputStream(outputStream.toByteArray)
+    val streamReader = new java.io.InputStreamReader(inputStream)
+
+    assertEquals(xml.toString, XML.load(streamReader).toString)
+  }
+
   @UnitTest
   def t0663 = {
     val src = scala.io.Source.fromString("<?xml version='1.0' encoding='UTF-8'?><feed/>")
--- a/shared/src/main/scala/scala/xml/XML.scala
+++ b/shared/src/main/scala/scala/xml/XML.scala
@@ -63,7 +63,7 @@ object XML extends XMLLoader[Elem] {
   val preserve = "preserve"
   val space = "space"
   val lang = "lang"
-  val encoding = "ISO-8859-1"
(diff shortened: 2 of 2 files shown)


ashawley:
Is it worthwhile to add a `scala.deprecated` annotation to the method?

Rogach:
@ashawley In that case we will need to create another method (`saveFile`, for example) and apply the change there. Is it worth it?

ashawley:
Oh, right. I didn't mean deprecate.  Sorry about that.  Is there a way to emit a warning that the behavior of the method has changed?

Rogach:
@ashawley - never saw anything like that. And since scala does not have any method for silencing warnings, I doubt there is one (else everybody would be stuck looking at that warning message on each compilation).

ashawley:
Ok, good points.  Maybe add a note in the method's documentation about the change then?  We'll have to notify the town crier as well.

Rogach:
Done. Who's the town crier?

SethTisue:
we can town-cry about it in the release notes for whatever Scala 2.12.x version bumps the module version

src/main/scala/scala/xml/XML.scala
```diff
@@ -74,6 +74,10 @@ object XML extends XMLLoader[Elem] {
    * Saves a node to a file with given filename using given encoding
    *  optionally with xmldecl and doctype declaration.
    *
+   *  Note: default encoding was ISO-8859-1 (latin1) in pre-1.0.7 scala-xml versions.
+   *  If your code depends on charaters in non-ASCII latin1 range, specify
```
SethTisue: s/charaters/characters/

Rogach: Shame on me :(

ashawley:
Unit test in another PR that will fail, but will be fixed by this PR.

Rogach:
@ashawley Should I include that test in this PR?

ashawley:
Sure.  Feel free to cherry-pick or rebase in here.

Rogach:
Is there something else I need to do to get this PR accepted?

SethTisue:
needs rebase now. LGTM otherwise and I would suggest we merge in a few days, after the rebase, if there are no objections

Rogach:
Done.

Seems rebase loses commit times, sadly (I'm more used to merge workflow).

SethTisue:
> Seems rebase loses commit times, sadly

are you sure? I don't think so

(regardless, thanks for taking care of it)

Rogach:
I see it now - my log viewer was set up to display commiter (`%cr`) date instead of author (`%ar`). I fixed it, thanks for pointing it out!

ashawley:
Still LGTM!

ashawley:
If you could do one more rebase and push, I'll merge this, tomorrow.

Rogach:
@ashawley - Done.

As a side note, I have huge troubles with rebase workflow - in this case, for example, I had to rebase the changes, thus rewriting history, and then had to do `push -f` to github (which is usually frowned upon). How rebase workflow is supposed to work in case of several distributed repositories?

ashawley:
Thanks for responding so quickly.  And apologies, for troubling you by repeatedly asking you to rebase.   Especially, if you aren't used to our git workflow.

Yes, rebase does rewrite git history on your branch.  It looks like your branch name for this PR is called `master`.  Is that the reason you have trouble?  Usually, I create a new branch called `utf-8-encoding` or `fix-121`. That way I can

```
git checkout master
git pull upstream master
git checkout fix-121
git rebase master
git push -f origin fix-121
```

Hope that helps!

Rogach:
Thanks for the explanation!

No, master branch does not give much trouble (though it will be a little bit more convenient if I did name the PR branch differently). But the thought of routinely doing `push -f` scares me a lot (and different branch name does not save from that). Merge workflow results in a little dirtier history, but saves me from messing with history rewriting.

ashawley:
Yeah, push force is weird.  In addition to `git reflog`, most of the git commands leave a trace in thehir output of what sha hashes you're moving between, so you can always recover or study them if you need to.  I am also paranoid, so I ocassionally inspect my commits and diffs after a rebase.  Although, over time and greater understanding you eventually learn to trust what rebase does, and git generally, more and more.

PR #133, opened 2017-02-04
Remove test duplication

OlivierBlanvillain:
This PR removes the test duplication introduced in #109.

As a defaults, new test should be added in `shared/src/test`. Parts that are JVM specific should go in `jvm/src/test` instead.


jvm/src/test/scala/scala/xml/XMLTest.scala
```diff
@@ -17,213 +17,14 @@ import java.io.StringReader
...
 
-object XMLTest {
+object XMLTestJVM {
   val e: scala.xml.MetaData = Null //Node.NoAttributes
   val sc: scala.xml.NamespaceBinding = TopScope
 }
 
-class XMLTest {
+class XMLTestJVM {
   import XMLTest.{ e, sc }
```
ashawley: `import XMLTestJVM`... although why there's a companion object, I have no idea.

OlivierBlanvillain: Good catch! Force pushed the change.

ashawley:
I don't think there's any reason not to merge this.  The duplication is indeed fixed.  I'm sure there will be more re-work in the wake of these Scalajs changes.

SethTisue:
I'm going to go ahead and hit "merge" on this one since it may affect in-flight PRs.

Issue #107, opened 2016-07-27
ConstructingParser adds space between two predefined entities when no other text is present


cayhorstmann:
scala.xml.parsing.ConstructingParser.fromSource(scala.io.Source.fromString("<div>&lt;&lt;</div>"), false).document
    res0: scala.xml.Document = Document(<div>&lt; &lt;</div>)

Note the space between the two `&lt;`.

cayhorstmann:
The problem is that predefined entities are turned into Atom(c), where c is the character represented by the entity. For example, &lt; turns into an Atom('<'). This happens in MarkupParser:

    case '&' => // EntityRef or CharRef
          nextch(); ch match {
            case '#' => // CharacterRef
              nextch()
              val theChar = handle.text(tmppos, xCharRef(() => ch, () => nextch()))
              xToken(';')
             ts &+ theChar

The &+ operator is defined in NodeBuffer as follows:

    def &+(o: Any): NodeBuffer = {
      o match {
        case null | _: Unit | Text("") => // ignore
        case it: Iterator[_]           => it foreach &+
        case n: Node                   => super.+=(n)
        case ns: Iterable[_]           => this &+ ns.iterator
        case ns: Array[_]              => this &+ ns.iterator
        case d                         => super.+=(new Atom(d))
      }
      ...
    }

That sounds innocuous, but there is no good way to *save* such a thing. Calling XML.write or PrettyPrinter.formatNodes ultimately goes through Utilities.sequenceToXML and this branch: 

    else if (children forall isAtomAndNotText) { // add space
        val it = children.iterator
        val f = it.next()
        serialize(f, pscope, sb, stripComments, decodeEntities, preserveWhitespace, minimizeTags)
        while (it.hasNext) {
          val x = it.next()
          sb.append(' ')
          serialize(x, pscope, sb, stripComments, decodeEntities, preserveWhitespace, minimizeTags)
        }
    }

When saving an element that contains only non-text atoms, they are separated by spaces.
The simplest remedy would be to change

    ts &+ theChar 

to

    ts &+ Text(theChar.toString)

in MarkupParser.

ashawley:
Good analysis of the problem.  Not sure if that change to `MarkupParser` will fix it though.

[...]

Issue #77, opened 2015-07-29
XML ConstructingParser too aggressive trimming whitespace around character references


mbeckerle:
This issue migrated from https://issues.scala-lang.org/browse/SI-3527.

Original description of the issue:

when preserveWS = false. example:
```
Welcome to Scala version 2.8.0.RC3 (Java HotSpot(TM) 64-Bit Server VM, Java 1.6.0_20).
 
scala> xml.parsing.ConstructingParser.fromSource(io.Source.fromString("<x>a &amp; b</x>"), preserveWS = false).document.text
res0: String = a&b
```
note that spaces on either side of the escaped ampersand have been lost.

I can't see why you would ever want this behavior. and I'm not an XML laywer, but section 4.4 of the XML 1.0 spec suggests that the character encoded by a character reference should be "retrieved and processed, in place of the reference itself, as though it were part of the document at the location the reference was recognized"

perhaps vaguely related: https://github.com/scala/scala-xml/issues/73

Issue #73, opened 2015-07-29
scala.xml.Utility.trim() doesn't properly handle adjacent Text nodes


mbeckerle:
This issue migrated from https://issues.scala-lang.org/browse/SI-3062. 

Please note: There is discussion there which is not reproduced here clarifying the needed algorithmic change.

Reproduced here is just the original description of the issue.

if `Text("My name is ")` followed by `Text("Harry")` the space following the word "is" will be incorrectly trimmed. Adjacent Text nodes need to be combined before whitespace is removed.

```
scala> import scala.xml._
import scala.xml._

scala> <div>{Text("My name is ")}{Text("Harry")}</div>
res0: scala.xml.Elem = <div>My name is Harry</div>

scala> Utility.trim(res0)
res1: scala.xml.Node = <div>My name isHarry</div>
```

This is important when modifying XML and then trimming it. For example we might start with `<div>My name is <user:name/></div>` and then replace the `<user:name/>` Elem with `"Harry"` thus leading to the adjacent Text nodes.

PR #140, opened 2017-04-02
Fix automatic space insertion between entity references in ConstructingParser. 

piyush-jaiswal:
This PR addresses the issue pointed out in [#107](https://github.com/scala/scala-xml/issues/107). The tests are the ones which were provided by @ashawley [here](https://github.com/ashawley/scala-xml/commit/a1ee9acbb7b3849ee8db1ff09e27cb9ab215037e). It would be great if this could be reviewed. Thanks!

### DIFF
--- a/shared/src/main/scala/scala/xml/Utility.scala
+++ b/shared/src/main/scala/scala/xml/Utility.scala
@@ -243,6 +243,12 @@ object Utility extends AnyRef with parsing.TokenTests {
       }
     }
 
+  // Checks if node when converted to string is an entity ref
+  def checkNodeForEntityRef(n: Node): Boolean = {
+    val st = n.toString 
+    st.startsWith("&") && st.endsWith(";")
+  }
+
   def sequenceToXML(
     children: Seq[Node],
     pscope: NamespaceBinding = TopScope,
@@ -256,10 +262,13 @@ object Utility extends AnyRef with parsing.TokenTests {
       else if (children forall isAtomAndNotText) { // add space
         val it = children.iterator
         val f = it.next()
+        var prev: Node = f
         serialize(f, pscope, sb, stripComments, decodeEntities, preserveWhitespace, minimizeTags)
         while (it.hasNext) {
           val x = it.next()
-          sb.append(' ')
+          // No need to append if space is between two EntityRefs. This is taken care in appendText in MarkupParser
+          if (!checkNodeForEntityRef(prev) && !checkNodeForEntityRef(x)) sb.append(' ')
+          prev = x
           serialize(x, pscope, sb, stripComments, decodeEntities, preserveWhitespace, minimizeTags)
         }
       } else children foreach { serialize(_, pscope, sb, stripComments, decodeEntities, preserveWhitespace, minimizeTags) }
--- a/shared/src/main/scala/scala/xml/parsing/MarkupParser.scala
+++ b/shared/src/main/scala/scala/xml/parsing/MarkupParser.scala
@@ -405,10 +405,19 @@ trait MarkupParser extends MarkupParserCommon with TokenTests {
   def appendText(pos: Int, ts: NodeBuffer, txt: String): Unit = {
     if (preserveWS)
       ts &+ handle.text(pos, txt)
-    else
-      for (t <- TextBuffer.fromString(txt).toText) {
-        ts &+ handle.text(pos, t.text)
+    else {
+      // If 'txt' is just made up of one or more spaces and 'ts' is not empty
+      if (!ts.isEmpty && TextBuffer.fromString(txt).toText == Nil) {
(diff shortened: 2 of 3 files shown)


SethTisue:
(sorry to nitpick as this looks like a nice improvement, but can you make sure the code you've added is indented correctly?)

piyush-jaiswal:
Yeah sure, no worries!

ashawley:
The tests might be failing because they need to go in the `XMLTest.scala` in `jvm/src/test/` instead of `shared/src/test`.   The test suite was a bit of a mess, and then it got more confusing when cross compilation to scalaJS was contributed.  My unit tests were before that happened.  I predict the `ConstructingParser` is not ScalaJS-friendly.

Would you be willing to try moving the tests and see if they pass?

piyush-jaiswal:
Yeah, I was trying to figure out the problem and found out that `java.io.File` is not portable to Scala.js. Thanks for pointing out the solution. Since the tests pass now, are there any further issues here?

ashawley:
It might not be necessary to introduce `Utility.checkNodeForEntityRef`.  You can probably just use `isAtom`, see http://www.scala-lang.org/api/current/scala-xml/scala/xml/Node.html#isAtom:Boolean

piyush-jaiswal:
Yeah, it is not necessary and does indeed work. Thanks!

shared/src/main/scala/scala/xml/Utility.scala
```diff
@@ -243,6 +243,12 @@ object Utility extends AnyRef with parsing.TokenTests {
       }
     }
 
+  // Checks if node when converted to string is an entity ref
+  def checkNodeForEntityRef(n: Node): Boolean = {
+    val st = n.toString
+    st(0) == '&' && st(st.length - 1) == ';'
```
SethTisue: do we know that `.toString` can never be empty?

also, trivial nitpick, but `st.last` is clearer than `st(st.length - 1)` imo

shared/src/main/scala/scala/xml/parsing/MarkupParser.scala
```diff
@@ -405,10 +405,19 @@ trait MarkupParser extends MarkupParserCommon with TokenTests {
...
     if (preserveWS)
       ts &+ handle.text(pos, txt)
-    else
-      for (t <- TextBuffer.fromString(txt).toText) {
-        ts &+ handle.text(pos, t.text)
+    else {
+      // If 'txt' is just made up of one or more spaces
+      if (TextBuffer.fromString(txt).toText == Nil) {
+        // Check if the last node in 'ts' was an 'Atom' and the next node to be parsed is an entity or character ref
+        if(ts(ts.length - 1).isAtom && ch == '&')
```
SethTisue: suggest: `ts.last`

SethTisue: do we know that `ts` can never be empty here?

shared/src/main/scala/scala/xml/Utility.scala
ashawley: @SethTisue I agree.  I'm hoping we can just get rid of this definition in favor of `isAtom`.  If not, then I was going to suggest using `startsWith` and `endsWith` to avoid the inevitable out of bounds errors here.

piyush-jaiswal: I'm not sure if we can completely get rid of `checkNodeForEntityRef`, as a test seems to fail when using 

`if(!prev.isAtom && !x.isAtom) sb.append(' ')`
instead of
`if (!checkNodeForEntityRef(prev) && !checkNodeForEntityRef(x)) sb.append(' ')`

1a7268b4-a67b-4ae7-8255-edf3af10abfb

I'm struggling at finding the alternatives to this definition.

ashawley: Ok, bummer.  Thanks for looking in to that.  I really appreciate it.

I guess I'll need to come up with some more test cases to better understand how the `ConstructingParser` parses entities.

piyush-jaiswal:
Hi, it's been some time since my last activity. I took a look at [#77](https://github.com/scala/scala-xml/issues/77) and found it somewhat relatable as it also involves whitespaces around text and entity refs. Hence, I've tried to fix the issue. Waiting for your review. Thanks!

ashawley:
> I'm not sure if we can completely get rid of checkNodeForEntityRef, as a test seems to fail when using `isAtom`

Ok, bummer. Thanks for looking in to that. I really appreciate it.

I guess I'll need to come up with some more test cases to better understand how the ConstructingParser parses entities.

I kind of wonder if #73 and #77 and #107 are all related problems. Optimistically, we choose a path to fix all of them, but if we can't, then choose which one(s) we can fix.

I'll try to study @piyush-jaiswal's fix for #77.  Thanks for looking in to that.

PR #145, opened 2017-04-25
Allow publishing existing tags against new Scala versions

lrytz:
This should include all changes proposed in https://github.com/scala/scala-xml/pull/141


ashawley:
@lrytz We're just waiting on a scalajs build for 2.13.0-M1?

lrytz:
Ah yes, it seems so. I heard this should happen soon, if there's nothing unexpected.

SethTisue:
Sebastien says 0.6.16 should be out later this week (0.6.15 cannot be published for 2.13)

lrytz:
This is ready now, the cross-build with 2.13.0-M1 passed.

Issue #154, opened 2017-08-02
NodeSeq.Empty is not serializable


joescii:
It appears that the typical use case of `NodeSeq` serializes with no problems (I'm speaking of `Node`, which we can see [here](https://github.com/scala/scala-xml/blob/master/shared/src/main/scala/scala/xml/Elem.scala#L90) is marked as `Serializable`), but the `NodeSeq.Empty` surprisingly doesn't. I would expect it to behave like `String` where the empty string is trivially serializable. 

Is there any reason that `NodeSeq` should not extend `Serializable`? The offender in the case of `NodeSeq.Empty` could trivially be marked to extend `Serializable` [here](https://github.com/scala/scala-xml/blob/master/shared/src/main/scala/scala/xml/NodeSeq.scala#L25)

You can see the behavior below in this console session. Note that java version details are printed at the end. This was run on a MBP running Sierra 10.12.6.

```text
[info] Starting scala interpreter...
[info] 
Welcome to Scala 2.12.3 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_131).
Type in expressions for evaluation. Or try :help.

scala> import java.io._
import java.io._

scala> import scala.xml.NodeSeq
import scala.xml.NodeSeq

scala>       def serialize(in: Any): Array[Byte] = {
     |         val bos = new ByteArrayOutputStream()
     |         val oos = new ObjectOutputStream(bos)
     |         oos.writeObject(in)
     |         oos.flush()
     |         bos.toByteArray()
     |       }
serialize: (in: Any)Array[Byte]

scala> serialize("joescii")
res0: Array[Byte] = Array(-84, -19, 0, 5, 116, 0, 7, 106, 111, 101, 115, 99, 105, 105)

scala> serialize("")
res3: Array[Byte] = Array(-84, -19, 0, 5, 116, 0, 0)

scala> serialize(<joescii/>)
[...]

PR #155, opened 2017-08-03
Nodeseq serialization

joescii:
Address issue #154. Includes tests for the JVM source.


joescii:
Oh nice! I love some scalacheck too. I'll subscribe to that particular thread so I can fill in serialization tests once it's ready.

PR #159, opened 2017-08-15
Reverting PR #150

joescii:
This reverts PR #150 due to [the discovery that it is a breaking change for Lift](https://github.com/scala/scala-xml/pull/150#issuecomment-319839436).


joescii:
Never mind this as a PR... this may not need to go to `master` depending on the branching strategy for breaking changes vs non-breaking changes

PR #179, opened 2017-12-24
add Java 9 test

### LINKED ISSUE #173: Add JDK 9 to Travis
Running in Java 9 should be possible.  It may require an update to sbt?

Trying it in Travis, I get the following:

    error: error while loading package, Missing dependency 'object java.lang.Object in compiler mirror',
    required by /home/travis/.sbt/boot/scala-2.10.6/lib/scala-library.jar(scala/package.class)
    error: error while loading package, Missing dependency 'object java.lang.Object in compiler mirror'
    required by /home/travis/.sbt/boot/scala-2.10.6/lib/scala-library.jar(scala/runtime/package.class)
    scala.reflect.internal.MissingRequirementError: object java.lang.Object in compiler mirror not found.

xuwei-k:
fix #173

### DIFF
--- a/admin/build.sh
+++ b/admin/build.sh
@@ -50,4 +50,4 @@ if [[ "$TRAVIS_TAG" =~ $tagPat ]]; then
   openssl aes-256-cbc -K $K -iv $IV -in admin/secring.asc.enc -out admin/secring.asc -d
 fi
 
-sbt "$publishVersion" "$publishScalaVersion" clean update +test +publishLocal $extraTarget
+sbt -Dsbt.scala.version=2.10.7 "$publishVersion" "$publishScalaVersion" clean update +test +publishLocal $extraTarget
--- a/build.sbt
+++ b/build.sbt
@@ -1,7 +1,7 @@
 import ScalaModulePlugin._
 
 scalaVersionsByJvm in ThisBuild := {
-  val v211 = "2.11.11"
+  val v211 = "2.11.12"
   val v212 = "2.12.3"
   val v213 = "2.13.0-M2"
   Map(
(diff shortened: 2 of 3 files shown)


admin/build.sh
```diff
@@ -50,4 +50,4 @@ if [[ "$TRAVIS_TAG" =~ $tagPat ]]; then
   openssl aes-256-cbc -K $K -iv $IV -in admin/secring.asc.enc -out admin/secring.asc -d
 fi
 
-sbt "$publishVersion" "$publishScalaVersion" clean update +test +publishLocal $extraTarget
+sbt -Dsbt.scala.version=2.10.7 "$publishVersion" "$publishScalaVersion" clean update +test +publishLocal $extraTarget
```
ashawley: I saw @SethTisue mention this, but I'd rather avoid breaking the build for people.

build.sbt
```diff
@@ -25,13 +25,23 @@ lazy val xml = crossProject.in(file("."))
...
+          .map(path =>
+            Map(
+              file(path
+                .split(java.io.File.pathSeparator)
+                .filter(_.endsWith(java.io.File.separator + "rt.jar")).head)
+              -> url("http://docs.oracle.com/javase/8/docs/api")
+            ))
+          .getOrElse(Map.empty)
+      }
+    }
```
ashawley: This refactoring actually causes an NPE to silently succeed with the `doc` task.  It's better if the task still fails, IMO.

This part of the sbt config may get replaced by https://github.com/ThoughtWorksInc/sbt-api-mappings

SethTisue:
waiting on https://github.com/scala/sbt-scala-module/issues/32

build.sbt
ashawley: Sorry, I take this back.  We will need to degrade when the Java boostrap classpath is unavailable.  I'll be using this PR as a basis for fixing Java 9 including the external doc linking.

ashawley:
Continuing in #180

Issue #121, opened 2016-11-08
ISO-8859-1 is used as default encoding


Rogach:
ISO-8859-1 (also known as latin1) is used as default encoding of `XML.save`:
https://github.com/scala/scala-xml/blob/master/src/main/scala/scala/xml/XML.scala#L67

This results in errors if document contains non-latin1 characters. Why this encoding is used by default instead of UTF-8?

ashawley:
> Why this encoding is used by default instead of UTF-8?

Presumably for backwards compatability.

In 2003, ISO-8859-1 became the default for `save` in 5d19bdaa.

Changing the default to UTF-8 would be a good improvement to make in a major release.

Would it help to add a mention of ISO-8859-1 to the documentation, with an example of using UTF-8?

```scala
    scala.xml.XML.save("foo.xml", <foo/>, "UTF-8")
```

ashawley:
That ISO-8859-1 was the default, was mentioned in Scala 2.6.1:

http://www.scala-lang.org/api/2.6.1/scala/xml/XML$object.html

Today, ISO-8859-1 is not mentioned:

http://scala-lang.org/files/archive/api/2.12.0/scala-xml/scala/xml/XML$.html

 It was removed in 31b4fe1.

Rogach:
I understand the value of backwards compatibility, but sadly the next major release would probably never happen.

Yes, adding a mention to documentation would definitely help! Actually, if value was displayed in actual function signature (`enc: String = "ISO-8859-1"`) it would be immediately obvious, but since it is factored out into value scaladoc only displays `enc: String = encoding`.

Should I make a pull request with this change?

P.S. "Another big XML commit" - and here I am, teaching people about value of good commit messages =)

SethTisue:
I'm not sure I agree that an actual fix would need to wait until the next major version

Rogach:
How else would it be possible? It would probably break code that depends on latin1 characters outside of ASCII range.

SethTisue:
I guess it depends how much meaning you attach to the version number. I'm not a semver purist, but some are. I would argue that user code that didn't specify an encoding had a latent bug in it and we should feel free to break that code, with a prominent explanation in the release note.

Basically I think we should ship a fix sooner rather than later. I don't have a strong opinion about the version number.

Rogach:
I, for one, am very much for it - my code would not break :)

ashawley:
The value of `encoding` in `save` is used in the XML declaration:

```xml
[...]

PR #203, opened 2018-04-05
Discard control characters in Utility.escape

hosamaly:
I encountered some input that happened to contain some control characters. This PR removes them from the output.

The previous behaviour used to exclude control characters in the smaller range. This PR excludes the ones between [`0x7F` (ASCII 127) and `0x9F` (ASCII 159)](https://docs.oracle.com/javase/10/docs/api/java/lang/Character.html#isISOControl(char)), plus possible future updates that Java supports.

### DIFF
--- a/shared/src/main/scala/scala/xml/Utility.scala
+++ b/shared/src/main/scala/scala/xml/Utility.scala
@@ -113,7 +113,7 @@ object Utility extends AnyRef with parsing.TokenTests {
     text.iterator.foldLeft(s) { (s, c) =>
       escMap.get(c) match {
         case Some(str)                             => s ++= str
-        case _ if c >= ' ' || "\n\r\t".contains(c) => s += c
+        case _ if !c.isControl || "\n\r\t".contains(c) => s += c
         case _ => s // noop
       }
     }


hosamaly:
@ashawley I'm using scala-xml v1.1.0. The previous behaviour is in relation to this PR.

@mbeckerle Thanks for the information. I'm not an expert here. My rationale is that control characters such as `0x7F`, `0x9D`, and `0x9F` are a security risk (although I'm not an expert there either).

I'm no longer sure about the best course of action. What do you recommend?

ashawley:
We should probably study the issue further with tests.  Unfortunately, we don't have good test coverage on `Utility.escape` in my opinion.  I'd be willing to accept a new pull request that documents the current behavior in unit tests.  After we merge it, we can better study your proposal, because we are friendly to your change in some form.  I will try to work on writing unit tests, but feel free to beat me to it.

hosamaly:
Thanks, @ashawley. Unfortunately, I didn't have time to work on this further but thanks for your PR. 👍

ashawley:
Ok, the new tests against `Utility.escape` show that the proposed change to `Utility.escape` causes only two differences:

```
[error] Test scala.xml.UtilityTest.escapeAsciiControlCharsTest failed: expected:< \n\r\t[^?]> but was:< \n\r\t[]>, took 0.042 sec
[error] Test scala.xml.UtilityTest.escapeUnicodeExtendedControlCodesTest failed: expected:<[]> but was:<[]>, took 0.004 sec
```

Above, `^?` is DEL (`\u007F`) and the`` is `\u0080`.

So this PR would effectively filter out `0x7F` from output as well as the range of obscure continuation bytes in UTF-8 (`0x80` to `0x9F`).  That's the extent of this PR.  It does exactly what the javadoc Hosam linked to for `java.lang.Character.isISOControl` says.  That seems acceptable to me, but it raises the concerns Mike Beckerle raised about the "[Characters](https://www.w3.org/TR/xml/#charsets)" section of the XML standard.

I suppose it also raises some of the same line of questions that were raised when we switched to the UTF-8 file encoding in #121.  In particular, can we assume every user of scala-xml works with a UTF-8 file encoding?  It seems like this change assumes so.  I don't know enough about Unicode to know whether there is a risk of regression for those who are in something other than UTF-8.

mbeckerle:
I believe this issue is unrelated to what character set encoding is being used.

The problematic C1 control codes that are the so called isoControl characters are characters, not bytes in some encoding. If a data file contains UTF-16 encoding, these control code simply occupy 16 bits instead of 8. They are Unicode-standard code points.

I believe we need to adhere to the XML standard here, and cannot make any character code, not even 7F illegal when the XML standard says it is legal. Apache Daffodil (incubator) a data parser that would definitely be broken hugely by such a change, and our users would encounter the issue whenever they try to parse data that contains these character codes and render it as XML. 

That said, I'm not opposed to an option where these C1 controls are treated the same as the C0 controls, (except for U+0085 which should always be allowed) it just can't be the default behavior, which I believe must strictly respect the XML standards.

ashawley:
Indeed, the XML spec says that XML "document authors" are "encouraged" to avoid these and other characters.  I suppose the compiler should be changed to avoid allowing them in XML literals. All other XML data that passes through this library is in the author's hands.

I spent some more time looking at this, and it might be possible to add an option that can avoid these "discouraged control characters".  We could probably add a parameter to `XML.write` or `PrettyPrinter.format`.

SethTisue:
I'm closing this since as per @ashawley's review, simply discarding the characters doesn't seem like the right path forward here. Perhaps a new PR could take up his suggestion about an alternate path.

PR #207, opened 2018-04-09
add -Xfuture option. fix procedure syntax warnings

xuwei-k:


ashawley:
Thanks for doing this, Kenji.  

Your commit message says "add -Xfuture option."  Did you forget to include a file?  Maybe build.sbt?

xuwei-k:
> Did you forget to include a file? Maybe build.sbt?

Oops. fixed

ashawley:
Hey @SethTisue, is `-Xfuture` something we can consider enabling or does it cause a time warp in either the community build universe or compiler bootstrap?  I'm thinking of a scenario where a new `-Xfuture` compiler error could keep the bootstrap process from working.  Maybe you could think of other scenarios.

SethTisue:
I would not suggest using `-Xfuture` except possibly as an alternate additional testing path — not by default. It isn't something you want enabled in a release build. (Not because of bootstrapping in particular, just general concerns about quality. You want to use the best-tested version of the compiler for doing release builds of core libraries.)

`-Xfuture` may go away anyway in 2.13: https://github.com/scala/scala-dev/issues/471

ashawley:
I'll merge, but undo `-Xfuture` and add a note per Seth's comments.

PR #215, opened 2018-05-09
Fix #214 Add threadsafe note for pretty printer

### LINKED ISSUE #214: PrettyPrinter is not threadsafe.
The PrettyPrinter, according [to the docs](https://www.scala-lang.org/api/2.11.7/scala-xml/#scala.xml.PrettyPrinter):

>Class for pretty printing. After instantiating, you can use the format() and formatNode() methods to convert XML to a formatted string. The class can be reused to pretty print any number of XML nodes.

The last sentence about being able to print any number of xml nodes I took to mean that the class operates in a pure manner and I could use this a concurrent environment. That was probably a bad assumption on my part. But here's a test case to reproduce the problem:

```
import scala.concurrent._
import scala.xml._
import java.util.concurrent.CountDownLatch
import scala.concurrent.ExecutionContext.Implicits.global

val p = new PrettyPrinter(width = Int.MaxValue, step = 2)
def xmlNode() = <A><B><C>Hello</C><D>There</D></B></A>
val latch = new CountDownLatch(1)
 
val futurePretty = (0 to 10).map { _ =>
	Future {
		val x = xmlNode()
		latch.await()
		p.format(x)
	}
}

latch.countDown()

val l = Future.sequence(futurePretty)

l.value.map(println)
```

On a run on my machine I see the result of that last console command as:

```
scala> l.value.map(println)
Success(Vector(<A>
  <B>
  <B>
    <C>Hello</C>
    <C>Hello</C>
    <C>Hello</C>
    <D>There</D>
  </B>
    <D>There</D>
  </B></B>
    <D>There</D>
</A>, <A>
  <B>
  <B>
    <C>Hello</C>
    <C>Hello</C>
... snipped for brevity...
```

Which clearly indicates to me that the pretty printer is using some buffer internal to the instance and therefore calling `.format` from different threads causes this buffer to be filled with values from each calling thread. This jumbles things up a bit. 

If it's unlikely that this issue will be fixed at any point, could we please update the documentation to leave an explicit note that the class should be used in this manner?

EdgeCaseBerg:
This is a minor documentation update to address #214

### DIFF
--- a/shared/src/main/scala/scala/xml/PrettyPrinter.scala
+++ b/shared/src/main/scala/scala/xml/PrettyPrinter.scala
@@ -21,6 +21,7 @@ import Utility.sbToString
  *  @author  Burak Emir
  *  @param width the width to fit the output into
  *  @param step  indentation
+ *  @note This class is not threadsafe and should not be accessed by multiple threads at the same time.
  */
 class PrettyPrinter(width: Int, step: Int, minimizeEmpty: Boolean) {
 


NthPortal:
Thanks for making this PR!

NthPortal:
by the way, if you could add the line "Fixes #214" to the end of the PR description, so that GitHub will automatically close the issue when this is merged, that would be great :)

ashawley:
> by the way, if you could add the line "Fixes #214" to the end of the PR description

Ethan put it in the PR title, so I think that works, as well.

NthPortal:
@ashawley oh oops, I did not realize that worked. Although it's a typo in the title and references the wrong issue

EdgeCaseBerg:
>Although it's a typo in the title and references the wrong issue

Woah, code catch. I updated the title of the PR!

PR #233, opened 2018-05-29
Fix #231: Minimize empty elements with lengthy attributes

### LINKED ISSUE #231: PrettyPrinter(minimizeEmpty = true) injects a new line into elements with attributes
Using scala-xml v1.1.0, pretty-printing a leaf node with attributes when `minimizeEmpty` is set to true injects a new line into the element:
```scala
scala> new scala.xml.PrettyPrinter(width = 4, step = 2, minimizeEmpty = true).format(<a b="c"/>)
res0: String =
"<a
b="c">
</a>
"

scala> scala.xml.XML.loadString(res0).child.map(_.getClass.getName)
res2: scala.collection.immutable.Seq[String] = List(scala.xml.Text)
```
This behaviour renders the XML invalid if the schema dictates a non-mixed element.

hosamaly:


hosamaly:
The build failure is unrelated to the changes in this PR.

jvm/src/test/scala/scala/xml/XMLTest.scala
```diff
@@ -749,4 +749,20 @@ class XMLTestJVM {
     assertEquals("<node>\n  <leaf/>\n</node>", pp.format(x))
   }
 
+  @UnitTest
+  def issue231: Unit = {
+    val pp = new xml.PrettyPrinter(4, 2, minimizeEmpty = true)
+    val x = <a b="c"/>
+    val formatted = pp.format(x)
+    assertEquals(x, XML.loadString(formatted))
```
ashawley: Shouldn't this test be comparing strings?  With `XML.loadString`, I think you're just verifying the XML structure is preserved from the original.

jvm/src/test/scala/scala/xml/XMLTest.scala
```diff
@@ -749,4 +749,20 @@ class XMLTestJVM {
...
+    assertTrue(formatted.trim.lines.length >= 2)
+  }
+
+  @UnitTest
+  def issue231_withoutAttributes: Unit = {
+    val pp = new xml.PrettyPrinter(4, 2, minimizeEmpty = true)
+    val x = <abcdefg/>
+    val formatted = pp.format(x)
+    assertEquals(x, XML.loadString(formatted))
```
ashawley: Same here.

jvm/src/test/scala/scala/xml/XMLTest.scala
hosamaly: I'm verifying that there are no extra nodes caused by the formatting, such as the `Text` node reported in #231.

I didn't want to hard-code a string to avoid having to modify the test if the formatting behaviour changes. For example, if the commented-out `todo` is fixed, the output may have 3 lines instead of 2.
```scala
// current behaviour
pp.format(x) == s"""<a\n  b="c"/>"""
// possible future behaviour
pp.format(x) == s"""<a\n  b="c"\n  />"""
```
Since the bug is about the XML structure changing (and, in my case, failing schema validation), I think that using `XML.loadString` is a better option.

What do you think?

middilman: Ah, ok.  That's a fair point.  The XML structure was getting changed.

There are other tests of `PrettyPrinter` that compare strings.  If history is any guide, I don't think it's a huge risk for the formatting behavior to change often.  Again, strings are the output of `PrettyPrint.format`, so it seems worth checking.

ashawley: Ah, ok.  That's a fair point.  The XML structure was getting changed.

There are other tests of `PrettyPrinter` that compare strings.  If history is any guide, I don't think it's a huge risk for the formatting behavior to change often.  Again, strings are the output of `PrettyPrint.format`, so it seems worth checking.

hosamaly: I'm not sure what value a string comparison would add. I think that the functionality of formatting is covered by other tests, e.g. `issue90` on line 746. This test also checks that some formatting happened by checking the number of lines in the output.

I'll add the string comparison if you deem it necessary, but I don't think it would add much value.

ashawley:
I may try to circle back and refactor this change and even add more tests, but it does fix the defect.  Thanks for doing this, Hosam!

PR #235, opened 2018-06-25
add 'oraclejdk10' to Travis build matrix

sullis:


sullis:
lgtm

ashawley:
Looks good to me.  Want to amend the commit and use the e-mail you use with GitHub?

```bash
$ git commit --amend --author 'Sean Sullivan <your@email>'
```

ashawley:
I've merged #229, so could you rebase with the latest master commits?

ashawley:
Since, Java 9 support already ended in March 2018, maybe we should replace Java 9 with Java 10? 
 Essentially, Java 9 is really just a milestone on the way to the long-term release version, Java 12.

SethTisue:
> maybe we should replace Java 9 with Java 10

eventually, for sure. whether it's necessary to do so now, no strong opinion

ashawley:
Unfortunately, the build count has gone from 12 to 20.  There's not really a big [binary or source compatibility change in Java 10](https://blogs.oracle.com/java-platform-group/introducing-java-se-10), unlike previous versions of Java where there was.  We could add more exclusion rules in the Travis config, but I'm just going to follow-up with switching Java 9 for Java 10.

Issue #256, opened 2018-08-30
publish for Scala 2.13.0-M5


SethTisue:

SethTisue:
probably blocked for now by https://github.com/scala-js/scala-js/issues/3443 ?

SethTisue:
"Scala.js 0.6.25 is now released with support for Scala 2.13.0-M5"

@ashawley is anything blocking us here? do you expect to get to this soon, like in the next couple days, say... or would you like the Scala team to handle it this time...? either way's fine

SethTisue:
if the answer is "we're blocked by https://github.com/scala/sbt-scala-module/issues/41", gah, what a mess, it seems there are multiple possible ways forward there and I've gotten lost in all the options

in this repo, at https://github.com/scala/scala-xml/issues/217, for M4 it looks like we didn't need to completely sort this out, instead we can just push a tag like `v1.1.0#2.13.0-M5#8` (after including any M5 changes) and it should work, it worked last time. yes? or am I missing something?

ashawley:
I updated my collectionsBootstrap branch, and got it to compile on M5.  I was just waiting on Scala.js, but it sounds like that's ready now.  I'll try to work on getting it publishing in the next few days.

Indeed, we were able to publish 2.13.0-M4 as you describe.  The Java 6 failures won't keep us from publishing 2.13.0-M5.

ashawley:
I pushed a new tag `v1.1.0#2.13.0-M5#8` and it looks like it published to sonatype.

I've updated the [newCollectionBootstrap](https://github.com/scala/scala-xml/compare/newCollectionsBootstrap).  However, it is broken because of:

- Serialization problem, #254 
- Scala.js 1.0.0-M3 not having a 2.13.0-M5 build

Previously, #217.

SethTisue:
@ashawley looking at https://travis-ci.org/scala/scala-xml/builds/428459277 and at the staging repos on Sonatype, it looks like you hit the same problem I did at https://github.com/scala/scala-parser-combinators/issues/170, where the Travis job for JDK 6 (https://travis-ci.org/scala/scala-xml/jobs/428459278) actually ran on JDK 8, so artifacts got staged twice. there are two staging repos, orgscala-lang-1607 and orgscala-lang-1608 that both contain similar (possibly identical?) artifacts

I think you need to do the equivalent of https://github.com/scala/scala-parser-combinators/pull/156 here. (or if we already did that here but in some other branch, cherry-pick the change)

[...]

Issue #254, opened 2018-08-30
Serialization in 2.13.0-M5


ashawley:
I tried compiling on the latest M5 with [Seth's branch](https://github.com/scalacommunitybuild/scala-xml/tree/community-build-2.13) and get serialization issues:

```
[error] Test scala.xml.SerializationTest.unmatched failed: expected:<> but was:<List()>, took 0.078 sec
[error] Test scala.xml.SerializationTest.xmlLiteral failed: expected:<<node/>> but was:<List(scala.collection.generic.DefaultSerializationProxy@28ec9fc9)>, took 0.013 sec
[error] Test scala.xml.SerializationTest.implicitConversion failed: expected:<<child></child><child/>> but was:<List(List(scala.collection.generic.DefaultSerializationProxy@5844dad8), List(scala.collection.generic.DefaultSerializationProxy@3204a005))>, took 0.002 sec
[error] Test scala.xml.SerializationTest.empty failed: expected:<> but was:<List()>, took 0.001 sec
[error] Test scala.xml.XMLTestJVM.serializeAttribute failed: java.lang.ClassCastException: scala.collection.immutable.$colon$colon cannot be cast to scala.xml.Attribute, took 0.011 sec
[error]     at scala.xml.XMLTestJVM.serializeAttribute(XMLTest.scala:218)
[error]     ...
[error] Test scala.xml.XMLTestJVM.serializeDocument failed: java.lang.ClassCastException: scala.collection.immutable.$colon$colon cannot be cast to scala.xml.Document, took 0.016 sec
[error]     at scala.xml.XMLTestJVM.serializeDocument(XMLTest.scala:228)
[error]     ...
[error] Test scala.xml.XMLTestJVM.serializeElem failed: java.lang.ClassCastException: scala.collection.immutable.$colon$colon cannot be cast to scala.xml.Elem, took 0.002 sec
[error]     at scala.xml.XMLTestJVM.serializeElem(XMLTest.scala:236)
[error]     ...
[error] Test scala.xml.XMLTestJVM.serializeComplex failed: java.lang.ClassCastException: scala.collection.immutable.$colon$colon cannot be cast to scala.xml.Elem, took 0.082 sec
[error]     at scala.xml.XMLTestJVM.serializeComplex(XMLTest.scala:271)
[error]     ...
[error] Failed: Total 164, Failed 8, Errors 0, Passed 156
[error] Failed tests:
[error] 	scala.xml.XMLTestJVM
[...]

PR #260, opened 2018-09-14
Move to Scala 2.13.0-M5

ashawley:
Previously, #222.

Work from this branch has already been published from the 1.1.0 tag in #256.

- [x] Serialization problem, #254, affects only 2.13 builds
- [x] Scala.js 1.0.0-M5 not having a 2.13.0-M5 build

### DIFF
--- a/shared/src/main/scala-2.11-2.12/scala/xml/ScalaVersionSpecific.scala
+++ b/shared/src/main/scala-2.11-2.12/scala/xml/ScalaVersionSpecific.scala
@@ -20,3 +20,7 @@ private[xml] trait ScalaVersionSpecificNodeSeq extends SeqLike[Node, NodeSeq] { 
 private[xml] trait ScalaVersionSpecificNodeBuffer { self: NodeBuffer =>
   override def stringPrefix: String = "NodeBuffer"
 }
+
+private[xml] trait ScalaVersionSpecificIterableSerializable[+A] { // extends Iterable[A] {
+  // protected[this] override def writeReplace(): AnyRef = this
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ jdk:
 scala:
   - 2.11.12
   - 2.12.6
-  - 2.13.0-M4
+  - 2.13.0-M5
 
 env:
   global:
@@ -46,7 +46,7 @@ matrix:
       jdk: openjdk11
     - scala: 2.12.6
       jdk: openjdk6
-    - scala: 2.13.0-M4
+    - scala: 2.13.0-M5
       jdk: openjdk6
 
 script: 
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,16 +30,16 @@ env:
     # The empty SCALAJS_VERSION will only compile for the JVM
     - SCALAJS_VERSION=
     - SCALAJS_VERSION=0.6.25
-    - SCALAJS_VERSION=1.0.0-M5
+    - SCALAJS_VERSION=1.0.0-M6
 
 matrix:
   exclude:
     - jdk: openjdk11
       env: SCALAJS_VERSION=0.6.25
     - jdk: openjdk11
-      env: SCALAJS_VERSION=1.0.0-M5
+      env: SCALAJS_VERSION=1.0.0-M6
     - scala: 2.11.12
-      env: SCALAJS_VERSION=1.0.0-M5
+      env: SCALAJS_VERSION=1.0.0-M6
     - scala: 2.11.12
       jdk: oraclejdk8
     - scala: 2.11.12
--- a/build.sbt
+++ b/build.sbt
@@ -19,7 +19,7 @@ lazy val xml = crossProject(JSPlatform, JVMPlatform)
     scalacOptions in Test  += "-Xxml:coalescing",
 
     mimaPreviousVersion := {
-      if (System.getenv("SCALAJS_VERSION") == "1.0.0-M5") None // No such release yet
+      if (System.getenv("SCALAJS_VERSION") == "1.0.0-M6") None // No such release yet
       else Some("1.1.1")
     },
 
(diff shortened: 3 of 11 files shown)


ashawley:
I had lost 33e5c85, but I've recovered it.  It's now here and rebased as eab3a79.

shared/src/main/scala-2.11-2.12/scala/xml/ScalaVersionSpecific.scala
```diff
@@ -0,0 +1,26 @@
...
+  override protected[this] def newBuilder = NodeSeq.newBuilder
+}
+
+private[xml] trait ScalaVersionSpecificNodeBuffer { self: NodeBuffer =>
+  override def stringPrefix: String = "NodeBuffer"
+}
+
+private[xml] trait ScalaVersionSpecificIterableSerializable[+A] { // extends Iterable[A] {
+  // protected[this] override def writeReplace(): AnyRef = this
+}
```
ashawley: Adding this null trait for 2.11 and 2.12 causes a strange error:

```
> xml/testOnly scala.xml.PatternMatching
[error] Test scala.xml.PatternMatching.nodeSeq failed: expected:<<a/><b/>> but was:<List(<a/>, <b/>)>, took 0.047 sec
[error] Failed: Total 6, Failed 1, Errors 0, Passed 5
```

ashawley:
I've merged the 2.13.0-M4 changes to master, in hopes of leaving only the changes for 2.13.0-M5 here.  Surprisingly, they don't look reduced here, but you can see the changes here:

https://github.com/scala/scala-xml/compare/newCollectionsBootstrap

SethTisue:
I think you'd need to rebase this one to see the reduced diffs

shared/src/main/scala-2.11-2.12/scala/xml/ScalaVersionSpecific.scala
xuwei-k: old `SafeNodeSeq.unapplySeq` returns `List`, but new `SafeNodeSeq.unapplySeq` returns `NodeSeq`.


`NodeSeq == List` always return `false`.

```
$ scala -language:_
Welcome to Scala 2.12.7 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_181).
Type in expressions for evaluation. Or try :help.

scala> import scala.xml._
import scala.xml._

scala> :paste
// Entering paste mode (ctrl-D to finish)

  object SafeNodeSeq {
    def unapplySeq(any: Any): Option[collection.immutable.Seq[Node]] = any match {
      case s: Seq[_] => Some((s flatMap (_ match {
        case n: Node => n case _ => NodeSeq.Empty
      })).toSeq) case _ => None
    }
  }

// Exiting paste mode, now interpreting.

defined object SafeNodeSeq

scala> SafeNodeSeq.unapplySeq(List(<a/>)).get.getClass.getSuperclass
res0: Class[?0] forSome { type ?0 >: T; type T <: scala.collection.immutable.Seq[scala.xml.Node] } = class scala.xml.NodeSeq

scala> :paste
// Entering paste mode (ctrl-D to finish)

  object SafeNodeSeq {
    def unapplySeq(any: Any): Option[Seq[Node]] = any match {
      case s: Seq[_] => Some(s flatMap (_ match {
        case n: Node => n case _ => NodeSeq.Empty
      })) case _ => None
    }
  }

// Exiting paste mode, now interpreting.

defined object SafeNodeSeq

scala> SafeNodeSeq.unapplySeq(List(<a/>)).get.getClass
res1: Class[_ <: Seq[scala.xml.Node]] = class scala.collection.immutable.$colon$colon
```

ashawley: Actually, it turns out that the trait has nothing to do with the issue.  Merging the 2.13.0-M4 fixes and leaving the 2.13.0-M5 changes helped to isolate the issue.

ashawley:
One thing that came up in the original collections rewrite changes for 2.13.0-M4 in #222 were the new  version-specific traits.  We made these traits `private`.  This access modifier means people can't use it as a trait.  However, the traits are published, and they are revealed in the type signatures.

scaladoc for scala.xml.NodeSeq

```
 :type -v <x y="z"/>.attributes
// Type signature
scala.xml.MetaData

// Internal Type structure
TypeRef(
  TypeSymbol(
    abstract class MetaData extends AbstractIterable[scala.xml.MetaData] 
                   with Iterable[scala.xml.MetaData] 
                   with Equality 
                   with ScalaVersionSpecificIterableSerializable[scala.xml.MetaData] 
                   with Serializable

  )
)
```

Fortunately (or unfortunately), dropping such a trait from a future version will be caught by Mima:

```
> xml/mimaReportBinaryIssues
[info] scala-xml: found 13 potential binary incompatibilities while checking 
[info] against org.scala-lang.modules:scala-xml_2.13.0-M5:1.1.1
...
[error]  * the type hierarchy of class scala.xml.NodeSeq is different in 
[error]  current version. Missing types {scala.xml.ScalaVersionSpecificIterableSerializable}
[error]    filter with: ProblemFilters.exclude[MissingTypesProblem]("scala.xml.NodeSeq")
```

Cost of doing business or should we do something else?  These weren't released in any of the 2.11 and 2.12 builds, yet, so I feel it's not too late to change course.

SethTisue:
> Cost of doing business or should we do something else?

gah, it's not _too_ horrible, but yuck anyway. do you have an alternative in mind?

SethTisue:
just MiMa complaining, now

ashawley:
> gah, it's not _too_ horrible, but yuck anyway. do you have an alternative in mind?

Well, I suppose we could source the entire files in the version-specific directories.  That would be the alternative, but it's not ideal.

SethTisue:
yay!

PR #265, opened 2018-09-26
use openjdk11 not oraclejdk11

SethTisue:
historically the Oracle one was the most standard one, but with the
new licensing model for 11, it's really OpenJDK that's the standard
one now, and the Oracle one is a commercial-use thing

sequel to #262

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ sudo: required
 jdk:
   - openjdk6
   - oraclejdk8
-  - oraclejdk11
+  - openjdk11
 
 scala:
   - 2.11.12


ashawley:
OpenJDK can't be used commercially?  You mean proprietary?

SethTisue:
OpenJDK can be used by anybody.

OracleJDK is restricted-usage unless you pay. "You may not: use the Programs for any data processing or any commercial, production, or internal business purposes other than developing, testing, prototyping, and demonstrating your Application"

see https://blog.joda.org/2018/09/do-not-fall-into-oracles-java-11-trap.html

ashawley:
Gotcha, one presumes Travis is paying Oracle?

SethTisue:
> Gotcha, one presumes Travis is paying Oracle?

And/or they figure they're just a middleman and that their users are the ones responsible for compliance? And/or they figure it's covered for everybody by the "development and testing" clause?

Here in this repo I figure that the "development and testing" clause would cover us if we wanted to include oraclejdk11 in the matrix. But I think it's actual better to do the testing on OpenJDK since that's the most-standard (and, I expect, in-practice-most-used) one.

PR #275, opened 2019-01-31
include correct license info in POM

SethTisue:


SethTisue:
credit to @Sciss for noticing

PR #303, opened 2019-04-23
Rename unmanaged source directories

Philippus:
Support was dropped for scala 2.11, but this was not reflected in the unmanaged source directory names yet. The naming in the PR is the naming suggested in https://github.com/scala/collection-strawman/wiki/FAQ#how-do-i-cross-build-my-project-against-scala-212-and-scala-213.


ashawley:
When the first 2.14 milestone drops, we'll rename the directories again?  Maybe they should be just be named `scala-2.12 and `scala-2.13`?

Philippus:
We could, but maybe we won't need scala 2.14 specific changes?

ashawley:
Scala 3 would suggest there will be more compatibility changes in 2.14, but we can just revisit this when 2.14 starts rolling out.

PR #312, opened 2019-05-22
Cross-compile with Dotty

smarter:
Tests that do not compile or failed with Dotty were moved to
scala-2.x directories and should be investigated at some point.


SethTisue:
btw, shouldn't scala-xml be in the Dotty community build? https://github.com/lampepfl/dotty/tree/master/community-build/community-projects

smarter:
Yes, that's the next step :).

ashawley:
That was my main concern.  Pinning this build to a certain version of Dotty compiler or the sbt-dotty plugin would otherwise hide problems, so I'm glad to hear there is a Community Build for Dotty.

ashawley:
I want to check that it works locally, and then I expect it's probably good to go.  I will try to find some time to play around with it.

PR #314, opened 2019-05-29
Dotty-related fixes and upgrades

smarter:


ashawley:
Thanks for fixing the REPL errors.

PR #315, opened 2019-05-29
support Scala 2.13.0-RC3 (drop RC2)

SethTisue:


ashawley:
We used to have a MiMa conditional for Scala.js 1.0.0-Mx, but it got removed in d3c6dd7, should be put back.

PR #334, opened 2019-07-08
Use AdoptOpenJDK JDK 8 and 11

eed3si9n:


.travis.yml
```diff
@@ -1,8 +1,8 @@
-language: scala
+sudo: false
+dist: xenial
```
SethTisue: can you add a comment explaining why this is needed?

also isn't `sudo:` redundant/unsupported these days?

eed3si9n: I want to explicitly specify the build environment, instead of Travis randomly picking one. It's probably true about `sudo`. I can remove that.

dwijnand: I'm not saying it's a bad idea, but what's the reason to pin the distribution?  Have you found issues in Travis CI upgrading builds to newer Ubuntu distributions and, if so, what types of issues?

I ask because we probably don't want to micro-manage nor lag in this configuration either.

lrytz: I also vote for minimal config, staying with defaults, unless there's actually an issue. If something breaks with the default config, it's their problem. If we have a non-standard config, it's ours.

eed3si9n: This is motivated by security as Ubuntu 14.04 LTS reached EOL on April 2019.
Travis CI did make xenial the default
https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment
but sometimes it takes a while for the rollout to reach older repos as these build image may contain breaking changes. 
See for example https://travis-ci.org/scala/scala/jobs/554521880 log where it's using trusty stable.

By picking xenial now, we can migrate our own choice of timing.

ashawley: That's not an issue here.  We're using Xenial by default.  I wonder if the Scala build is using Trusty because the Travis config has `sudo: required` in it?

eed3si9n: I just wanted to make sure we're using xenial, so if that's the default Travis CI is already using then I'm happy to remove the specification.

eed3si9n: For what it's worth, scala/scala defaulted to trusty without `sudo` - https://travis-ci.org/scala/scala/jobs/556437704

.travis.yml
```diff
@@ -19,20 +15,30 @@ env:
...
+before_install:
+  # adding $HOME/.sdkman to cache would create an empty directory, which interferes with the initial installation
+  - "[[ -d /home/travis/.sdkman/ ]] && [[ -d /home/travis/.sdkman/bin/ ]] || rm -rf /home/travis/.sdkman/"
+  - curl -sL https://get.sdkman.io | bash
+  - echo sdkman_auto_answer=true > /home/travis/.sdkman/etc/config
+  - source "/home/travis/.sdkman/bin/sdkman-init.sh"
+
+install:
+  - sdk install java $(sdk list java | grep -o "$ADOPTOPENJDK\.[0-9\.]*hs-adpt" | head -1)
+  - unset _JAVA_OPTIONS
```
dwijnand:
```suggestion
```

.travis.yml
```diff
@@ -19,20 +15,30 @@ env:
...
-    - scala: 0.15.0-RC1
-      env: SCALAJS_VERSION=1.0.0-M8
+    - scala: 0.16.0-RC3
+      env: SCALAJS_VERSION=0.6.28   ADOPTOPENJDK=8
+    - scala: 0.16.0-RC3
+      env: SCALAJS_VERSION=1.0.0-M8 ADOPTOPENJDK=8
+
+before_install:
+  # adding $HOME/.sdkman to cache would create an empty directory, which interferes with the initial installation
+  - "[[ -d /home/travis/.sdkman/ ]] && [[ -d /home/travis/.sdkman/bin/ ]] || rm -rf /home/travis/.sdkman/"
```
dwijnand:
```suggestion
  - "[[ -d "$HOME/.sdkman/bin" ]] || rm -rf "$HOME/.sdkman/"
```

.travis.yml
```diff
@@ -19,20 +15,30 @@ env:
...
+  - curl -sL https://get.sdkman.io | bash
+  - echo sdkman_auto_answer=true > /home/travis/.sdkman/etc/config
+  - source "/home/travis/.sdkman/bin/sdkman-init.sh"
```
dwijnand:
```suggestion
  - source "$HOME/.sdkman/bin/sdkman-init.sh"
```

.travis.yml
```diff
@@ -19,20 +15,30 @@ env:
...
+    - scala: 0.16.0-RC3
+      env: SCALAJS_VERSION=0.6.28   ADOPTOPENJDK=8
+    - scala: 0.16.0-RC3
+      env: SCALAJS_VERSION=1.0.0-M8 ADOPTOPENJDK=8
+
+before_install:
+  # adding $HOME/.sdkman to cache would create an empty directory, which interferes with the initial installation
+  - "[[ -d /home/travis/.sdkman/ ]] && [[ -d /home/travis/.sdkman/bin/ ]] || rm -rf /home/travis/.sdkman/"
+  - curl -sL https://get.sdkman.io | bash
+  - echo sdkman_auto_answer=true > /home/travis/.sdkman/etc/config
```
dwijnand:
```suggestion
  - echo sdkman_auto_answer=true > "$HOME/.sdkman/etc/config"
```

.travis.yml
```diff
@@ -19,20 +15,30 @@ env:
...
+  - source "/home/travis/.sdkman/bin/sdkman-init.sh"
+
+install:
+  - sdk install java $(sdk list java | grep -o "$ADOPTOPENJDK\.[0-9\.]*hs-adpt" | head -1)
+  - unset _JAVA_OPTIONS
+  - unset JAVA_HOME
```
lrytz: Could you explain this line? Thanks!

eed3si9n: JAVA_HOME is set by Travis CI's base image or `language: scala` to be whatevet the JDK, like JDK 11 for xenial. since I'm installing a new JDK I don't want the java on PATH and JAVA_HOME to be not inconsistent. This normally only matters when there's forking or Java only project. Alternatively I could try to do `which java` and set JAVA_HOME to wherever SDKMAN installs AdoptOpenJDK JDK.

lrytz: OK, so `sdk install` adds the installed version to the path, but doesn't set `JAVA_HOME`. I'm fine with unsetting if we don't need it. For the record, `jabba use` sets both the `PATH` and `JAVA_HOME`.

eed3si9n:
Squashed.

eed3si9n:
@lrytz lol. good catch. I saw those too but somehow I didn't put 2 and 2 together.

I guess I'd use adoptopenjdk/openjdk8 or eed3si9n/sbt:jdk8-alpine Docker image.

eed3si9n:
@ashawley I am not familiar with CircleCI, but I am guessing that creating a custom image of a drop-in replacement of the current image using AdoptOpenJDK is going to be some work, so what do you think about switching back to Travis CI for now?

ashawley:
I haven't touched TravisCI.  It should still be enabled for the repo. Is there an issue with Travis?  Is there a syntax error in your yaml file? The travis UI changed recently and you have to find those errors from the requests page from under the hamburger menu.

eed3si9n:
Interesting. We noticed it's no longer running, but maybe you're right about it potentially being an YAML error etc. The last known build on Travis CI was from this PR yesterday - https://travis-ci.org/scala/scala-xml/builds/556385195/config

.travis.yml
eed3si9n: Double quote broke YAML.

eed3si9n:
@ashawley you were totally right about YAML syntax.

eed3si9n:
Looks like it ran on Travis CI fine (in addition to Circle CI), so I am going to self-merge.

ashawley:
Yeah, I've ever been bit by both the YAML syntax problem, and the hidden error problem in Travis.

lrytz:
👍  funny episode

PR #374, opened 2019-12-03
Import travis config from scala-dev

lrytz:


lrytz:
@ashawley this reverts most of your changes from https://github.com/scala/scala-xml/pull/370, but I'd really like to keep all scala modules having the same configuration. Here's the shared config: https://github.com/scala/scala-dev/blob/scala-dev/travis/default.yml

PR #397, opened 2020-01-22
copyright 2020

SethTisue:


SethTisue:
(note that I didn't update the headers in the individual source files, e.g.:

```
shared/src/main/scala/scala/xml/Atom.scala:**    / __/ __// _ | / /  / _ |    (c) 2003-2019, LAMP/EPFL             **
```

it isn't necessary to have an up-to-date year in every file, and if we want to have one anyway, the sane way to do it would be with sbt-header)

Issue #257, opened 2018-09-05
Subsequent RewriteRules don't transform elements added in previous transform.


ndelargy:
````scala
import scala.xml._
import scala.xml.transform.{RewriteRule, RuleTransformer}

object TransformIssue {
  def addNewElement(): RewriteRule = new RewriteRule {
    override def transform(n: Node): Seq[Node] = n match {
      case <element></element> => <element><new></new></element>
    }
  }

  def addThingElement(): RewriteRule = new RewriteRule {
    override def transform(n: Node): Seq[Node] = n match {
      case <element>{ children@_*}</element> => <element>{ children  }</element>
      case <new></new> => <new><thing></thing></new>
    }
  }

  def change(node: Node): Node =
    new RuleTransformer(
      addNewElement(),
      addThingElement()
    ).transform(node).head

  def changeWorkaround(node: Node): Node = {
    val out1 = new RuleTransformer(
      addNewElement()
    ).transform(node).head

    new RuleTransformer(
      addThingElement()
    ).transform(out1).head
  }

}
````

````scala
import org.scalatest.{FlatSpec, FunSpec}
import org.scalatest._

class TransformIssueSpec extends FlatSpec with Matchers {

  it should "apply transform to created elements" in {
    val output = TransformIssue.change(<element></element>)
    output should be(<element><new><thing></thing></new></element>)
  } // fails

  it should "work the same as the workaround imo" in {
    TransformIssue.change(<element></element>) should equal(TransformIssue.changeWorkaround(<element></element>))
  } // fails

}
````

When we apply a transform with two rewrite rules: the first one adding a new element, the second one adding children to the new element; then the second rewrite rule does not match on the elements added in the first rule.

When we apply the same RewriteRules in two separate RuleTransformers it does add the children to the elements added in the first step.  We would expect the `change` and `changeWorkaround` functions to produce the same output.

**Comment by ashawley** [comment]
Indeed, this is a known issue.  On the [Getting started](https://github.com/scala/scala-xml/wiki/Getting-started) page, it is written:

> Keep in mind that rewrite rules won't compose if they modify children, or modify values that other rewrite rules depend on.

There's a chance that `RuleTransformer` executes rewrite rules backward.  What if you try changing the order?

```scala
    new RuleTransformer(
      addThingElement(),
      addNewElement()
    )
```

If that's not it, then indeed, it's probably that you need to call the transformer on the new elements yourself:

```scala
...
[...]

PR #421, opened 2020-03-21
Make RuleTransformer fully recursive [#257]

dcsobral:
RuleTransformer for the past eleven years or more has first recursed,
then applied the rewrite rules as it backed out of the recursion.

That prevented rewrite rules from acting on changes of previous rules
unless they recursed themselves.

The workaround has always been chain rule transformers instead of
calling one rule transformer with all the rules. This change basically
re-implements RuleTransformer as that workaround, and introduces
a NestingTransformer which does the nesting.

Clearly, if you have N rules you'll now recurse N times instead of
once, though each rule is still only applied once for each element.

On the other hand, a RewriteRule that recursed would incur in
exponential times, and now they don't need to.

The original behavior has no reason to be. It didn't prevent
rules from seeing each other changes, nor was it particularly
concerned with performance.  With API changes coming on 2.0, I
believe this is the right time to introduce this change.

### DIFF
--- a/shared/src/main/scala/scala/xml/transform/NestingTransformer.scala
+++ b/shared/src/main/scala/scala/xml/transform/NestingTransformer.scala
@@ -12,10 +12,8 @@ package transform
 
 import scala.collection.Seq
 
-class RuleTransformer(rules: RewriteRule*) extends BasicTransformer {
-  private val transformers = rules.map(new NestingTransformer(_))
+class NestingTransformer(rule: RewriteRule) extends BasicTransformer {
   override def transform(n: Node): Seq[Node] = {
-    if (transformers.isEmpty) n
-    else transformers.tail.foldLeft(transformers.head.transform(n)) { (res, transformer) => transformer.transform(res) }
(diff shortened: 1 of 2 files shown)


dcsobral:
Fixes #257

SethTisue:
is there someone watching this repo who could review this...?

dcsobral:
> is there someone watching this repo who could review this...?

And there was a deafening silence...

SethTisue:
@ashawley @dcsobral wdyt, should we maybe just merge this?

dcsobral:
@SethTisue I rely on the tests. I've always found it hard to reason about the XML transformation code.

SethTisue (approved):
yeah. seems fair game to me for 2.0, but let's give Aaron a week or so to object, perhaps he knows something you and I don't

dcsobral:
It's been 18 days now.

SethTisue:
another backstop is the community build; I've included this change here: https://github.com/scala/community-build/pull/1333

PR #441, opened 2020-09-06
Fix Invalid Comment Edge Case

isomarcte:
According to [the XML specification](https://www.w3.org/TR/xml11//#IDA5CES "W3 XML") XML comments can not end in `--->`. Prior to this commit creating a comment of the form `Comment("invalid-")` would yield no error though it should, since it is rendered as `<!--invalid--->`. This commit makes such `String` values yield an `IllegalArgumentException`.

A companion object was also added with two helper methods, `validated` and `validatedOpt`, which allow for a user of the library to check if a `String` is valid without having to `catch` the `IllegalArgumentException`.

### DIFF
--- a/shared/src/main/scala/scala/xml/Comment.scala
+++ b/shared/src/main/scala/scala/xml/Comment.scala
@@ -14,6 +14,8 @@ package xml
  *
  * @author Burak Emir
  * @param commentText the text contained in this node, may not contain "--"
+ *        and the final character may not be `-` to prevent a closing span of `-->`
+ *        which is invalid. [[https://www.w3.org/TR/xml11//#IDA5CES]]
  */
 case class Comment(commentText: String) extends SpecialNode {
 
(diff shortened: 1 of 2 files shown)


isomarcte:
I had to have the `Comment` companion object extend `scala.runtime.AbstractFunction1` in order to not break binary compatibility. 

I'm also fine changing that to any of the following,
* Filter the binary compatibility issue since it is unlikely anyone is depending on that.
* Bumping the version.
* Removing the companion object and the `Either`/`Option` based validation code. Though this is my least preferred option. As a user of this library, I'd personally prefer to have the validation code. That being said, I'm happy to defer to the maintainers' preference.

lrytz (commented):
Thank you for that patch!

I took a quick look around the codebase, and there are many other constructors that perform validation. I think adding the new validation methods only for `Comment` and leaving everything else makes the library less consistent. So my preference would be to only add the check in the constructor, but no new public API. Once it's there it has to stay for bin compat...

isomarcte:
@lrytz I've updated the PR and removed the companion object.

lrytz (approved):
Thanks!

PR #443, opened 2020-09-07
Drop support for scala.js 0.6.x and update scala.js to 1.1.1

### LINKED ISSUE #430: Switch to Scala.js 1.x
Make it the default in the build, or maybe scala-xml 2.0 will be a good opportunity to just remove 0.6.x from the build.

Philippus:
closes #430


SethTisue:
I'm going to go ahead and merge this, since I don't think there's any whiff of controversy on master. (Maybe there wouldn't be on the 1.x branch either.)

PR #470, opened 2020-11-26
Update For Scala 3.0.0-M2

isomarcte:


SethTisue:
thanks! I added some more commits bringing the build up to date in some other respects.

Issue #469, opened 2020-11-23
Publish API documentation


Jasper-M:
Looks like the API for the most recent version didn't get published.

https://scala.github.io/scala-xml/api/1.3.0/scala/xml/

SethTisue:
@ashawley should we ditch the GitHub.io based publication and just direct people to https://javadoc.io/doc/org.scala-lang.modules/scala-xml_2.13/1.3.0/index.html ? (did we discuss this before?)

ashawley:
I've published 1.3.0 just for completeness.  Yeah, feel free to ditch it.  Just need to stop publishing the apiUrl in `build.sbt` to the scala.github.io address in the 2.0.0-Mx milestone series.

SethTisue:
#472

PR #472, opened 2020-11-26
let javadoc.io host our Scaladoc

SethTisue:
as per discussion on #469


SethTisue:
Aaron already said he was okay with this, so, merging.

Issue #282, opened 2019-03-17
Scala Native Support


sadhen:

SethTisue:
I believe a pull request on this would be welcome, even if it only supported XML generation and not XML parsing (see https://github.com/scala/scala-xml/issues/198 for discussion on that)

SethTisue:
we've now dropped Scala 2.11 support here in this repo, so I think this would now depend on Scala Native adding support for 2.12

sadhen:
Scala Native 0.4.0 is released! Come back to see if there is a way to have a native scala-xml.

david-bouyssie:
I have ported the missing pieces for a full XML support on SN:
https://github.com/david-bouyssie/xml4s

The following artifact contains a pure Scala Pull Parser (fork of Kxml2), with an adapter for SAX:
https://search.maven.org/artifact/com.github.david-bouyssie/xml4s-xmlpull-api_native0.4_2.13/0.1.0/jar

On the long term it should also solve this similar Scala.js issue https://github.com/scala/scala-xml/issues/198

This work is still in early stage, so feel free to ask for improvements.

PR #482, opened 2021-01-23
Add support for Scala Native 0.4

lolgab:
Fixes: #282


lolgab:
@SethTisue It sounds right to me and the tests pass, but let's ask @WojciechMazur if he can :)

build.sbt
```diff
@@ -163,3 +163,11 @@ lazy val xml = crossProject(JSPlatform, JVMPlatform)
     Test / fork := false
   )
   .jsConfigure(_.enablePlugins(ScalaJSJUnitPlugin))
+  .nativeSettings(
+    scalaModuleMimaPreviousVersion := None, // No such release yet
+    // Scala Native cannot run forked tests
+    Test / fork := false,
+    Test / libraryDependencies += "org.scala-native" %%% "junit-runtime" % nativeVersion,
```
WojciechMazur, line 170: This should fix the compile time errors. Everything other LGTM, although I'm courious were this test should be finally placed. CircleCI has also configuration for ScalaJs, but not for ScalaNative
```suggestion
    Test / libraryDependencies += "org.scala-native" %%% "junit-runtime" % nativeVersion,
    libraryDependencies += "com.novocode"  % "junit-interface" % "0.11" % Test,
```

lolgab: Good to know! Let me try!

lolgab: @WojciechMazur Nope, it doesn't work if I move the dependencies to the Test scope, even adding `"com.novocode"  % "junit-interface" % "0.11" % Test`

WojciechMazur: Ok, so the problem is we're adding  native junit compilert plugin to the whole project, so even if we try to execute compilaton on standard sources it still will be trigered. 
We forget to add something similar to `ScalaJSJUnitPlugin` that would add compiler junit plugin only when compiling tests. 
I'm not sure, but probably for now the only option of adding `scala-native-junit-compiler-plugin` would be to manually add it to classpath ([here's how it's done in ScalaJs](https://github.com/scala-js/scala-js/blob/master/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSJUnitPlugin.scala) ).

sjrd: It should just be
```suggestion
    libraryDependencies += "org.scala-native" %%% "junit-runtime" % nativeVersion % Test,
```
`Test / libraryDependencies` is never used by sbt.

sjrd: But yes, it's true that the plugin should only be added in the `Test` configuration as well. This is not a feature that sbt offers by default. It can be done like we do in Scala.js, with a custom config. Or since sbt 1.x it's also possible to directly use the `librarymanagement` API, which might be easier. It should look something like
```scala
val lm = dependencyResolution.value
val cp = lm
  .retrieve("org.scala-native" % "junit-plugin_${scalaVersion.value}" % scalaNativeVersion, scalaModuleInfo = None, retrieveDir, log)
  .fold(w => throw w.resolveException, Attributed.blankSeq(_))
cp.find(... ~ "junit-plugin")
```

lolgab: @sjrd It doesn't work either. I get:
```
[error] scala.reflect.internal.MissingRequirementError: object org.junit.Test in compiler mirror not found.
```
@WojciechMazur I guess it's fine to depend on it in the compile scope ( it's what other libraries using JUnit are doing ) and change once the issue is fixed in a future Scala Native version?

sjrd: > I guess it's fine to depend on it in the compile scope

If you do that, then everyone who depends on your library gets JUnit on their classpath. We should instead enable the plugin only in the Test config.

lolgab: @sjrd Tried what you suggested. Can I ask you another look? :)

build.sbt
```diff
@@ -163,3 +163,27 @@ lazy val xml = crossProject(JSPlatform, JVMPlatform)
...
+      val cp = lm
+        .retrieve(
+          "org.scala-native" % s"junit-plugin_${scalaVersion.value}" % nativeVersion,
+          scalaModuleInfo = None,
+          retrieveDir,
+          log
+        )
+        .fold(w => throw w.resolveException, identity(_))
+      val jarPath = cp
+        .find(_.toString.contains(s"/org/scala-native/junit-plugin_${scalaVersion.value}/$nativeVersion/junit-plugin_${scalaVersion.value}-$nativeVersion.jar"))
```
sjrd, line 184: Not sure this will work on Windows. Since this classpath is going to contain very few stuff (junit-plugin, scala-compiler, scala-reflect and scala-library, basically), you can make this much easier:
```suggestion
        .find(_.toString.contains("junit-plugin"))
```
and that will work on any OS.

sjrd (approved):

SethTisue:
Would it be hard to also update the CircleCI config?

I don't really know where we stand with CircleCI. @ashawley added it as "experiment" and "backup" at https://github.com/scala/scala-xml/pull/324  Personally, I'm fine with Travis-CI, or perhaps GitHub Actions in the future. Normally having two CI systems around doesn't really cost us anything.... except times when it does, like now.

lolgab:
@SethTisue Added a CircleCI job to test on Scala Native

SethTisue:
thank you!!

SethTisue:
as for publishing it, https://github.com/scala/scala-xml/issues/484

.scalafmt.conf
```diff
@@ -0,0 +1 @@
+version = "2.7.4"
```
xuwei-k, line 1: https://github.com/scala/scala-xml/pull/582#issuecomment-1003230966 🤔

build.sbt
xuwei-k: - https://github.com/scala-native/scala-native/issues/2540

PR #487, opened 2021-02-01
copyright 2021

SethTisue:


Philippus (changes requested):
The original starting year (2002/2003) is lost in a lot of files.
We could also change the headers to lose the years, like we did it in https://github.com/scala/scala-parser-combinators/pull/184.

SethTisue:
> The original starting year (2002/2003) is lost in a lot of files.

Ha, oops.

> We could also change the headers in the source files to lose the years

Yeah, I'll just do that.

SethTisue:
I was able to get motivated to add sbt-header, but not sufficiently motivated to enforce it in CI

PR #501, opened 2021-03-12
add sbt-version-policy and advance mimaPreviousVersion

### LINKED ISSUE #497: Build should declare a `versionScheme`
as per https://scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html and discussion at scala/sbt-scala-module#111

SethTisue:
fixes #497


SethTisue:
Travis-CI is failing, but only on 2.12+JDK15 and 2.13+JDK15, with 

```
[error] scala-xml: Failed binary compatibility check against org.scala-lang.modules:scala-xml_2.13:2.0.0-M5! Found 1 potential problems
[error]  * method declaration(java.lang.String,java.lang.String,java.lang.String)Unit in class scala.xml.include.sax.XIncluder does not have a correspondent in other version
[error]    filter with: ProblemFilters.exclude[DirectMissingMethodProblem]("scala.xml.include.sax.XIncluder.declaration")
```

?!

in addition to looking into this, I should also update the CircleCI config to perform the same check that Travis-CI is

SethTisue:
note that it's the forward-compatibility check that is failing, not the backward one:

```
sbt:scala-xml> xml/mimaReportBinaryIssues
[success] Total time: 0 s, completed Mar 13, 2021, 11:25:00 AM
sbt:scala-xml> xml/versionPolicyCheck
[error] scala-xml: Failed binary compatibility check against org.scala-lang.modules:scala-xml_2.13:2.0.0-M5! Found 1 potential problems
...
```

so I assume the root cause here is that on the newer JDK version, there is a new method in one of `XIncluder`'s parents

I'll just add it to the filters.

SethTisue:
This is ready for review.

project/plugins.sbt
```diff
@@ -11,3 +11,4 @@ addSbtPlugin("org.scala-js" % "sbt-scalajs" % scalaJSVersion)
 addSbtPlugin("org.scala-native" % "sbt-scala-native" % scalaNativeVersion)
 addSbtPlugin("ch.epfl.lamp" % "sbt-dotty" % "0.5.3")
 addSbtPlugin("de.heikoseeberger" % "sbt-header" % "5.6.0")
+addSbtPlugin("ch.epfl.scala" % "sbt-version-policy" % "1.0.0-RC5")
```
lrytz, line 14: should we add this to https://github.com/scala/sbt-scala-module/blob/2.x/build.sbt instead?

SethTisue: That was consider and discussed at https://github.com/scala/sbt-scala-module/issues/111 . I was on the fence about it, but in the end I just shrugged and made a decision not to do it.

In general, my experiences while doing community build work have given me a somewhat dim view of putting build stuff in plugins. It makes individual builds harder to read and harder to maintain. As soon as settings and tasks are coming from a plugin, you can't just read the build and see what's going on anymore. You have to consider the possibility that a bunch of random stuff is coming in from some plugin and then you have to go read the sources of the plugin, too.

And then maintenance is harder for individual module maintainers, because they can't easily make changes to sbt-scala-module themselves. (And it's already hard enough to find and retain module maintainers.)

SethTisue: I suppose sbt-scala-module could add sbt-version-policy but not actually set the settings. That's a middle ground I'm not sure I considered.

But anyway, 🤷

SethTisue: Note that now that we have the Scala Steward, we get version bumps that way, so there's less benefit in getting them via sbt-scala-module.

PR #587, opened 2022-01-07
Handle XML namespace declarations even when the underlying parser is …

### LINKED ISSUE #506: Do not assume that parser is namespace-unaware

dubinsky:
…namespace-aware.

fixes #506

### DIFF
--- a/jvm/src/test/scala/scala/xml/XMLTest.scala
+++ b/jvm/src/test/scala/scala/xml/XMLTest.scala
@@ -629,6 +629,35 @@ class XMLTestJVM {
     // does not work: roundtripNodes("<a/> <!-- epilogue -->")
   }
 
+  // using non-namespace-aware parser, this always worked;
+  // using namespace-aware parser, this works with FactoryAdapter enhanced to handle startPrefixMapping() events;
+  // see https://github.com/scala/scala-xml/issues/506
+  def roundtrip(namespaceAware: Boolean, xml: String): Unit = {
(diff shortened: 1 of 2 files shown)


dubinsky:
@ashawley @SethTisue could you please take a look? thanks!

dubinsky:
@ashawley @SethTisue Is there anything I need to change in this pull request for it to get merged? Thanks!

SethTisue:
I know next to nothing about this codebase, so I'm unable to offer substantive review here.

Let's give Aaron a little more time, like a week or so, and if he doesn't turn up, I'd be happy to optimistically merge it, since you seem to be about as expert on this stuff as anybody.

dubinsky:
> Let's give Aaron a little more time, like a week or so, and if he doesn't turn up, I'd be happy to optimistically merge it, since you seem to be about as expert on this stuff as anybody.

You are too kind! I'll wait. I won't be able to use the new functionality anyway until it is released, and Aron wanted to follow up on my previous pull request with [few amendments to the API choices in this around naming](https://github.com/scala/scala-xml/pull/560#issuecomment-992641527)...

dubinsky:
@ashawley @SethTisue Is there anything I need to change in this pull request for it to get merged? Thanks!

PR #615, opened 2022-07-12
drop support for Scala 3.0.x

SethTisue:
I see this happening ecosystem-wide, we might as well

PR #616, opened 2022-07-28
fix Scala 3 publishing

SethTisue:
sequel to #615


SethTisue:
who knows if it's perfect yet, but maybe it's at least enough to straighten out https://github.com/lampepfl/dotty/pull/15774

PR #620, opened 2022-09-18
Set `apiURL`

armanbilge:


armanbilge:
Hm, since CI needs approval I would appreciate a ping so I know to come back and check/fix.

Or ... do us all a favor an in repo settings under "actions" set it so only users new to GH need approval, not every first-time contributor to this project :)

SethTisue (approved):

PR #648, opened 2023-02-07
Add type annotations

dubinsky:
Added type annotations for `val`s, `var`s and `def`s that did not have them; primary motivation: clarity.
Types inferred by IDEs like IntelliJ Idea are not always the types inferred by the Scala compiler.
In fact, different _versions_ of the Scala compiler infer types _differently_!
As a result, in around 20 places it is impossible to add missing type annotations without breaking binary compatibility...

### DIFF
--- a/jvm/src/test/scala-2.x/scala/xml/CompilerErrors.scala
+++ b/jvm/src/test/scala-2.x/scala/xml/CompilerErrors.scala
@@ -4,10 +4,10 @@ import org.junit.Test
 
 class CompilerErrors extends CompilerTesting {
   @Test
-  def t7185() = {
+  def t7185(): Unit = {
     // the error message here differs a bit by Scala version
     import util.Properties.versionNumberString
-    val thing =
+    val thing: String =
       if (versionNumberString.startsWith("2.12")) "method value"
       else "method"
     expectXmlError(s"""|overloaded $thing apply with alternatives:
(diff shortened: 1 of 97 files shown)


SethTisue:
This goes farther than I would personally go in a codebase I worked heavily on; something like e.g. `def foo: Boolean = false` seems like noise to me versus `def foo = false`. And a lot of the changes are to local variables, which I'm far less likely to want type annotations on than I am in method signatures.

Regardless, I don't actually object to this being merged if you think it's a win overall.

Oh, heh, you hit "merge" right as I was writing this comment :-)

dubinsky:
> This goes farther than I would personally go in a codebase I worked heavily on; something like e.g. `def foo: Boolean = false` seems like noise to me versus `def foo = false`. And a lot of the changes are to local variables, which I'm far less likely to want type annotations on than I am in method signatures.
> 
> Regardless, I don't actually object to this being merged if you think it's a win overall.
> 
> Oh, heh, you hit "merge" right as I was writing this comment :-)

Oops :(

You are right, some of the type annotations I added are trivial, but I wanted everything explicit because in some cases what I think the type is, what the IDE thinks and what the compiler thinks are all different...

SethTisue:
It's fine, suit yourself!

dubinsky:
Thanks!

PR #651, opened 2023-03-15
Fix infinite loop caused by unclosed character ref

### LINKED ISSUE #306: Infinite loop caused by unclosed character ref in MarkupParser#content
Putting an unclosed character ref (`&#`) in XML causes an infinite loop.
Example code:

```
import scala.io.Source
import scala.xml.parsing.XhtmlParser
val text = "<html><body>test &# test </body></html>"
XhtmlParser(Source.fromString(text))
```

mike-mcgann:
Resolves #306

If parseCharRef is called for an unclosed character reference it will get caught in an infinite loop if the end of file is reached. Add a check to exit the loop when an EOF is encountered.

### DIFF
--- a/shared/src/test/scala/scala/xml/UtilityTest.scala
+++ b/shared/src/test/scala/scala/xml/UtilityTest.scala
@@ -219,4 +219,15 @@ class UtilityTest {
     val x: Elem = <div>{Text("   My name ")}{Text("  is  ")}{Text("  Harry   ")}</div>
     assertEquals(<div>My name is Harry</div>, Utility.trim(x))
   }
+
+  @Test
+  def issue306InvalidUnclosedEntity(): Unit = {
+    val entity = "&# test </body></html>"
+    val it: Iterator[Char] = entity.iterator
+    var c = it.next()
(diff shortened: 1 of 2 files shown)


shared/src/test/scala/scala/xml/UtilityTest.scala
```diff
@@ -219,4 +219,15 @@ class UtilityTest {
     val x: Elem = <div>{Text("   My name ")}{Text("  is  ")}{Text("  Harry   ")}</div>
     assertEquals(<div>My name is Harry</div>, Utility.trim(x))
   }
+
+  @Test
+  def issue306InvalidUnclosedEntity(): Unit = {
```
dubinsky, line 224: Maybe the name of the test method should be more specific: this is about character references, not entities in general...

dubinsky (changes requested):
Once this merges, #306 will close; there is a comment there about a [different](https://github.com/scala/scala-xml/issues/306#issuecomment-1199769172) infinite loop - would you mind extracting it into a separate issue so that it is not lost? Thank you!

dubinsky (approved):
Thank you!!!

PR #655, opened 2023-03-27
Missing type annotations

dubinsky:
- introduce Scala-version-specific  MiMa exclusions
- add type annotations that break binary compatibility except for the `unapply()` methods for the `Node` subclasses;
- add appropriate MiMa exclusions

@SethTisue as discussed in #649:

> could we have one set of MiMa exclusions for Scala 2 and one for Scala 3, so that we can know for sure that any potential bincompat breakage is only on the Scala 3 side?

done; only Scala 3-specific exclusions are added in this pull request
**NOTE**: no longer true; I added all the remaining missing type annotations except for the `unapply()` methods: since even with the Scala 3-only exclusions this pull request is considered too risky (although not even one example of the breakage it can cause is known), I think it is better to record all the related changes in one pull request.

> do you understand why MiMa considers the more specific types to break binary compatibility?

probably because MiMa can not take subtyping into account since JVM class linking does not...

I understand the reluctance to break binary compatibility between the current and the next versions of this library _just_ to add some type annotations, but please consider that _because_ those type annotations are not there, binary compatibility is _already broken_ between the builds of the same version of this library for different versions of Scala: code that uses the methods with the type annotations missing that is built with Scala 2 will stop compiling - or will start behaving differently - when switched to Scala 3 _with the same version of this library_...

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -67,11 +67,42 @@ lazy val xml = crossProject(JSPlatform, JVMPlatform, NativePlatform)
     mimaBinaryIssueFilters ++= {
       import com.typesafe.tools.mima.core._
       import com.typesafe.tools.mima.core.ProblemFilters._
-      Seq(
+      Seq( // exclusions for all Scala versions
         // necessitated by the introduction of new abstract methods in FactoryAdapter:
         exclude[ReversedMissingMethodProblem]("scala.xml.parsing.FactoryAdapter.createComment"),  // see #549
-        exclude[ReversedMissingMethodProblem]("scala.xml.parsing.FactoryAdapter.createPCData")    // see #558
+        exclude[ReversedMissingMethodProblem]("scala.xml.parsing.FactoryAdapter.createPCData"),    // see #558
+        exclude[IncompatibleResultTypeProblem]("scala.xml.Node.NoAttributes"),
+        exclude[IncompatibleResultTypeProblem]("scala.xml.dtd.PublicID.attribute")
+      ) ++ (CrossVersion.partialVersion(scalaVersion.value) match {
+        case Some((3, _)) => Seq( // Scala 3-specific exclusions
+          exclude[IncompatibleResultTypeProblem]("scala.xml.Null.getNamespace"),
+          exclude[IncompatibleResultTypeProblem]("scala.xml.Null.next"),
+          exclude[IncompatibleResultTypeProblem]("scala.xml.Null.key"),
+          exclude[IncompatibleResultTypeProblem]("scala.xml.Null.value"),
+          // not sure why `Null.value` and `Null.apply` need `DirectAbstractMethodProblem`
+          // exclusion for `MetaData` but `Null.key` etc. do not
+          exclude[DirectAbstractMethodProblem]("scala.xml.MetaData.value"),
+          exclude[IncompatibleResultTypeProblem]("scala.xml.Null.remove"),
+          exclude[IncompatibleResultTypeProblem]("scala.xml.Null.apply"),
+          exclude[DirectAbstractMethodProblem]("scala.xml.MetaData.apply"),
+          exclude[IncompatibleResultTypeProblem]("scala.xml.dtd.NoExternalID.publicId"),
+          exclude[IncompatibleResultTypeProblem]("scala.xml.dtd.NoExternalID.systemId"),
+          // for `SpecialNode.child: Nil.type` and `Group.child: Nothing`:
+          exclude[DirectAbstractMethodProblem]("scala.xml.Node.child")
+        )
+        case Some((2, 13)) => Seq( // Scala 2.13-specific exclusions
+          exclude[IncompatibleResultTypeProblem]("scala.xml.Null.filter"),
+          exclude[IncompatibleResultTypeProblem]("scala.xml.Null.apply"),
+          exclude[DirectAbstractMethodProblem]("scala.xml.MetaData.apply")
+        )
+        case Some((2, 12)) => Seq( // Scala 2.12-specific exclusions
+          exclude[IncompatibleResultTypeProblem]("scala.xml.Null.filter"),
+          exclude[IncompatibleResultTypeProblem] ("scala.xml.Null.apply"),
+          exclude[DirectAbstractMethodProblem] ("scala.xml.MetaData.apply")
       )
+        case _ => Seq(
+        )
+      })
     },
     // Mima signature checking stopped working after 3.0.2 upgrade, see #557
     mimaReportSignatureProblems := (CrossVersion.partialVersion(scalaVersion.value) match {
--- a/shared/src/main/scala/scala/xml/Group.scala
+++ b/shared/src/main/scala/scala/xml/Group.scala
@@ -44,6 +44,6 @@ final case class Group(nodes: Seq[Node]) extends Node {
   override def label: Nothing = fail("label")
   override def attributes: Nothing = fail("attributes")
   override def namespace: Nothing = fail("namespace")
-  override def child /* TODO type annotation */ = fail("child")
+  override def child: Nothing = fail("child")
   def buildString(sb: StringBuilder): Nothing = fail("toString(StringBuilder)")
 }
--- a/shared/src/main/scala/scala/xml/Node.scala
+++ b/shared/src/main/scala/scala/xml/Node.scala
@@ -23,12 +23,12 @@ import scala.collection.Seq
  */
 object Node {
   /** the constant empty attribute sequence */
-  final def NoAttributes: MetaData = Null
+  final def NoAttributes: Null.type = Null
 
   /** the empty namespace */
   val EmptyNamespace: String = ""
 
-  def unapplySeq(n: Node) /* TODO type annotation */ = Some((n.label, n.attributes, n.child.toSeq))
+  def unapplySeq(n: Node) /* TODO type annotation Some[(String, MetaData, Seq[Node])] */ = Some((n.label, n.attributes, n.child.toSeq))
 }
 
 /**
--- a/shared/src/main/scala/scala/xml/Null.scala
+++ b/shared/src/main/scala/scala/xml/Null.scala
@@ -27,15 +27,15 @@ case object Null extends MetaData {
   override def iterator: Iterator[Nothing] = Iterator.empty
   override def size: Int = 0
   override def append(m: MetaData, scope: NamespaceBinding = TopScope): MetaData = m
-  override def filter(f: MetaData => Boolean): MetaData = this
+  override def filter(f: MetaData => Boolean): Null.type = this
 
   override def copy(next: MetaData): MetaData = next
-  override def getNamespace(owner: Node) /* TODO type annotation */ = null
+  override def getNamespace(owner: Node): scala.Null = null
 
   override def hasNext: Boolean = false
-  override def next /* TODO type annotation */ = null
-  override def key /* TODO type annotation */ = null
-  override def value /* TODO type annotation */ = null
+  override def next: scala.Null = null
+  override def key: scala.Null = null
+  override def value: scala.Null = null
   override def isPrefixed: Boolean = false
 
   override def length: Int = 0
@@ -47,8 +47,9 @@ case object Null extends MetaData {
   }
   override protected def basisForHashCode: Seq[Any] = Nil
 
-  override def apply(namespace: String, scope: NamespaceBinding, key: String) /* TODO type annotation */ = null
-  override def apply(key: String) /* TODO type annotation */ =
+  override def apply(namespace: String, scope: NamespaceBinding, key: String): scala.Null = null
+
+  override def apply(key: String): scala.Null =
     if (Utility.isNameStart(key.head)) null
     else throw new IllegalArgumentException("not a valid attribute name '" + key + "', so can never match !")
 
@@ -61,6 +62,6 @@ case object Null extends MetaData {
 
   override def wellformed(scope: NamespaceBinding): Boolean = true
 
-  override def remove(key: String) /* TODO type annotation */ = this
-  override def remove(namespace: String, scope: NamespaceBinding, key: String) /* TODO type annotation */ = this
+  override def remove(key: String): Null.type = this
+  override def remove(namespace: String, scope: NamespaceBinding, key: String): Null.type = this
 }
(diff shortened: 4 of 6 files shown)


SethTisue:
One possibility to try to give us some hope that downstream breakage wouldn't be widespread and immediate would be to try this out in the Scala 3 community build — not the big, “open” community build maintained by VirtusLab, which only tests varying the Scala 3 version, but the smaller community build in the lampepfl/dotty repo. If I understand correctly, in that smaller build an updated scala-xml would be used by downstream libraries.

SethTisue:
Another thing that could be help would be to put out an RC — 2.2.0-RC1, I guess it would be — and ask some important downstream users to try it out.

SethTisue:
While I'm throwing random ideas out there, here's one more:

We could declare a version 3 but move all the code to a new package, so that the new version and the old version could coexist on the same classpath and there also wouldn't be a riot of eviction warnings in everyone's builds, ecosystem-wide, as we all only just got (mostly) done doing with the 1->2 upgrade.

:sigh:

cc @ckipp01

ckipp01:
> as we all only just got (mostly) done doing with the 1->2 upgrade.

> :sigh:

> cc @ckipp01

_shivvers_

I don't really have many thoughts on this. I just know that if we knowing break compat again people will probably lose their minds. The update to 2.x was a way bigger deal than I thought it would be.

Since we are throwing wild ideas out there, if we were to break, coinciding with he sbt 2.x release could be the best time.... but again, I really really think we should not break again in the same manner that we did before.

ashawley:
The collections library update for 2.13 caused some changes for scala-xml, but they were deemed compatible enough and to only require a 1.3.0 release.  The 2.0 release series of scala-xml was intended to help land any significant changes for Scala 3 "dotty".  Those changes never came. Most of the changes for Scala 3 were required upstream in dotty. Unlike the Scala 2.13 effort, I never studied the dotty artifacts closely.  It's interesting that we missed these signature differences between the two archives.  Not a lot of people probably interface with these low-level bits of of scala-xml to notice.  It seems fine to try fixing these type signatures as a bug fix.  If the experiment fails we can rollback.

dubinsky:
@SethTisue 
>> probably because MiMa can not take subtyping into account since JVM class linking does not

> Seems like a good general description, but I was hoping we could understand the possible failure scenarios more precisely in order to gauge the danger level here. We can be (very nearly) sure that MiMa is technically correct here that failure scenarios exist, but I'm wondering how far-fetched they are. For example, do they only involve users extending scala-xml classes that are rarely or never extended in practice? Or is there a possibility of more ordinary usage of the library running into trouble?

> I'm asking not only in order to help us make an up/down decision about this PR, but also because even in the case of an "up" decision, we'll need to document the situation in the release notes.

I suspect that since method invocation on JVM references a [method descriptor](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.3.3) that specifies the method's return type, ordinary usage of the methods affected by this change _will_ cause breakage. I'll try to tabulate some of the failure scenarios illustrating this in the coming days...

@ashawley 
> Unlike the Scala 2.13 effort, I never studied the dotty artifacts closely. It's interesting that we missed these signature differences between the two archives. Not a lot of people probably interface with these low-level bits of of scala-xml to notice.

It's easy to miss these signature differences without either
- an automated binary compatibility check across the Scala versions for the same version of the library (which we do not have)
- or - better! - having all methods annotated with the return types, so that there _are_ no signature differences ;)

> The collections library update for 2.13 caused some changes for scala-xml, but they were deemed compatible enough and to only require a 1.3.0 release.

There is a bunch of `unapply()` methods which return a `Seq[Node]` as a part of their result, and since they do not have type annotations, their signatures are inferred by the Scala compiler, resulting in the signatures being different not only between Scala 2.13 and Scala 3, but _also between Scala 2.12 and Scala 2.13_. Those differences were missed too, and they _affect the run-time behavior of pattern-matching on `Node` subclasses_  - something that a lot of people probable _do_ interface with.

@ashawley 
> It seems fine to try fixing these type signatures as a bug fix. If the experiment fails we can rollback.

@ckipp01
> I really really think we should not break again in the same manner that we did before.

I am going to:
- look into the specifics of the breakage resulting from this change
- create another pull request that covers the `unapply()` methods (and binary compatibility breakage for Scala 2) 
so we can have a full(er) picture.

dubinsky:
@SethTisue @ckipp01  @ashawley my findings so far:

Proposed change affects return type of all the involved methods in a uniform way: 
- in the companion object, return type changes to a more specific one;
- in the class itself, a variant of the method with the same parameters but more specific return type appears alongside  what existed before the change. For example:
```
public static scala.xml.MetaData next(); // before the change
public static scala.runtime.Null$ next(); // after the change

public scala.xml.MetaData next(); // before and after the change
public scala.runtime.Null$ next(); // appears after the change
```

Reflection on the method continues to return the non-specific variant even after the change, contrary to what I expected after reading [this](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Class.java#L2272) comment on the `java.lang.Class.getMethod()`:
>The result of this algorithm is chosen arbitrarily from the methods with most specific return type among all selected methods...
There may be more than one method with matching name and parameter types in a class because while the Java language forbids a class to declare multiple methods with the same signature but different return types, the Java virtual machine does not. This increased flexibility in the virtual machine can be used to implement various language features.  For example, covariant returns can be implemented with bridge methods; the bridge method and the overriding method would have the same signature but different return types. This method would return the overriding method as it would have a more specific return type.

Turns I out, when I said the "MiMa can not take subtyping into account _since JVM class linking does not_", I was wrong - at least about the JVM: even reflection takes subtyping into account, let alone [invokevirual](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.10.1.9.invokevirtual) JVM instruction (being _virtual_ and all :)).  As a result, the code that uses `scala-xml-2.1.0` with Scala 3, where the compiler infers less specific types, and invokes a method affected by this change with such pre-change less specific type - for instance ```val next: MetaData = Null.next``` -  works _just fine_ with the version of `scala-xml` where the type changes to a more specific one (`scala.Null`), notwithstanding the fact that the less specific type used is encoded in the classfile (```invokevirtual  // Method scala/xml/Null$.next:()Lscala/xml/MetaData;```).

So far I was not able to see any breakage introduced by this change, nor figure out what breakage MiMa is warning about from its [code](https://github.com/lightbend/mima/blob/main/core/src/main/scala/com/typesafe/tools/mima/lib/analyze/method/MethodChecker.scala). It is clear that that code was written by people who understand this stuff much better than me (to put it mildly), but it is possible that the potential breakage MiMa is worried about materializes only in situations where different methods are called depending on how the calling code "sees" the class or some such(https://github.com/scala/bug/issues/11804, https://github.com/scala/scala/pull/9141) - concerns that does not seem to apply to this change.

@SethTisue 
> I was hoping we could understand the possible failure scenarios more precisely in order to gauge the danger level here

Maybe you know somebody to consult on this? It would be a shame to not apply this change, since unlike the breakage it _could_ cause, we _do_ know what existing breakage its absence _does_ cause: the code that uses `scala-xml-2.1.0` with Scala 2.12 or 2.13, where the compiler infers _more_ specific types, and invokes a method with such more specific type - for instance ```val next: scala.Null = Null.next``` - _does not work_ with `scala-xml-2.1.0` with Scala 3,  where the compiler infers _less_ specific types (`scala.xml.MetaData`): ```java.lang.NoSuchMethodError: 'scala.runtime.Null$ scala.xml.Null$.getNamespace(scala.xml.Node)'```; it doesn't even _compile_ using `scala-xml-2.1.0` with Scala 3 either (```Found: xml.MetaData, Required: Null```), forcing the developers facing this breakage to weaken the types...

In the next pull request in this series, the one addressing the `unapply()` methods, pre-existing breakage involved is more far-reaching (it affects pattern-matching on `Node` subclasses) and more insidious (silent code behavior changes); it would be nice to be able to take care of that...

SethTisue:
@dwijnand can we get your input here?

dubinsky:
@SethTisue do you think there is a way to get some input on this?

lrytz:
@SethTisue and me discussed this today.

IMO breaking binary compatibility of Scala 3 artifacts over this is not worth the risk (or the effort to release a new major version of xml).

To get explicit return types in the source, you can add a single top-level object with type aliases, with separate sources for 2 and 3. Something like this:

```scala
// For Scala 2
private[xml] object ReturnTypes {
  type NullNext = scala.Null
  type NullKey = scala.Null
  type NullValue = scala.Null
  ...
}
```


```scala
// For Scala 3
private[xml] object ReturnTypes {
  type NullNext = MetaData
  type NullKey = String
  type NullValue = Seq[Node]
  ...
}
```


and then change the definitions to use these aliases

```scala
case object Null extends MetaData {
...
  override def next: NullNext = null
  override def key: NullKey = null
  override def value: NullValue = null
...
}
```

ashawley:
I wonder if we determine we can't make these signature fixes or any binary changes in 2.0 really, since it is too risky, maybe we need to just fork the 2.0 branch and make a 3.0 RC version for Leonid and others to make progress with? Or make a forked library and repo, call it "scala-xml3" or such?

dubinsky:
@lrytz 
> To get explicit return types in the source, you can ...

Thank you for your suggestion; it allows to make the status-quo _explicit_, but not to make types to be the _same_ for Scala 2 and Scala 3...  Also, the cost of just making the current types explicit - declaring them indirectly via `ReturnTypes`  - makes it not worth it IMO.

@SethTisue 
>> code that uses the methods with the type annotations missing built with Scala 2 will stop compiling with Scala 3

> I would consider that source compatibility, not binary compatibility,

You are right, of course :)

@SethTisue 
> I don't think we should be bothered by minor source incompatibilities between the Scala 2 and Scala 3 versions. We might have avoided it if we'd seen it coming, but now that it's happened, I don't think we should feel bothered. Users haven't complained.
> Also, I think the Scala 3 types are actually fine. Having different types in 2 and 3 is a bit awkward, but I don't think the Scala 3 types are actually bad — the less specific types seem fine to me.

Maybe the users do not complain because in the midst of the migration from Scala 2 to Scala 3, minor changes like this are the least of their problems ;)

Forcing the users that are trying to use the most specific types available to weaken them bothers me, and so does having  different types for Scala 2 and Scala 3 for no good reason.

@SethTisue 
> When something doesn't even compile, that's actually good! The compiler is letting the user know there's a problem they need to address. It's far, far less pernicious than runtime breakage.

Absolutely! In this case, the problem is spurious though :(

@SethTisue 
> I'm not 100% sure I follow what you're saying about `NoSuchMethodError`, though. That's a runtime error, so that worries me.

Worry not: it happens when the Scala XML library built for Scala 2 is swapped for the one built for Scala 3 without recompiling the code using it - a scenario not covered by any compatibility guarantees.

@lrytz 
> IMO breaking binary compatibility of Scala 3 artifacts over this is not worth the risk

@SethTisue 
> do you understand why MiMa considers the more specific types to break binary compatibility?
> We can be (very nearly) sure that MiMa is technically correct here that failure scenarios exist, but I'm wondering how far-fetched they are. 

Is there a way to gauge this risk? So far, we are not aware of any scenarios where those changes lead to breakage...

@ashawley 
> I wonder if we determine we can't make these signature fixes or any binary changes in 2.0 really, since it is too risky, maybe we need to just fork the 2.0 branch and make a 3.0 RC version for Leonid and others to make progress with?

I think we are at that point, yes: I have more radical changes planned, and it would be nice to have a place for them; on the other hand, if chances of those changes ever being released are slim (as seems to be the case) - is there any point?

lrytz:
@dubinsky I apologize I missed your reply.

> it allows to make the status-quo _explicit_, but not to make types to be the _same_ for Scala 2 and Scala 3

I understand, and I believe this is the right thing to do here.

Of course it's not ideal to have a mismatch between the signatures of a method in Scala 2 vs 3. But:
  - the other solutions (breaking binary compatibility, or releasing scala-xml 3.0) are risky (binary compatibility is very hard to reason about, we failed more than once in the past) or very costly
  - we are talking about a couple of overrides that narrow to `Null` / `Nil` / `Nothing`. User code will not depend on these signatures, both the Scala 2 and Scala 3 versions will work equally well in ordinary user code (people are not going to write `Null.next`). It doesn't complicate cross-building / source compatibility.

> the cost of just making the current types explicit - declaring them indirectly via `ReturnTypes` - makes it not worth it IMO.

Why not? Result types of public methods should be explicit, as we are just witnessing here.

dubinsky:
> @dubinsky I apologize I missed your reply.

No problem @lrytz , thank you for the clarifications!

> > the cost of just making the current types explicit - declaring them indirectly via `ReturnTypes` - makes it not worth it IMO.

> Why not? Result types of public methods should be explicit, as we are just witnessing here.

Type annotations serve to 1) make the intent known to the compiler and 2) to document the types for the developer. What we are witnessing here is that result types of the public methods should _have been_ explicit to begin with to make the intent clear and avoid the not ideal signature mismatch between Scala versions. Making explicit the accidental status quo breaks 1) by codifying the situation that is not aligned with the intent, and `ReturnTypes.NullNext` is more confusing than `scala.Null`. 

Above is the reasoning behind my reluctance to make the types explicit but version specific; now that I understand that there is no chance to remedy the situation and roll the accident back, this reluctance is gone: *any* documentation of the types is better than no documentation... So, I shall make a pull request that follows your suggestion and add type annotations wherever they are missing. Thank you!!

By the way, signature mismatch is not limited to obscure things like `Null.next`; here are the signatures of a few methods of the class/object `scala.xml.Elem` (all `scala.xml.Node` subclasses are similarly affected):
```
// all Scala versions:
def copy(String, String, MetaData, NamespaceBinding, boolean, scala.collection.Seq[Node]): Elem

// Scala 2.12:
def apply(String, String, MetaData, NamespaceBinding, boolean, scala.collection.Seq[Node]): Elem
def unapplySeq(Node): Option[(String, String, MetaData, NamespaceBinding, scala.collection.Seq[Node])]
def child(): scala.collection.Seq[Node]

// Scala 2.13 & 3 (you can pass into copy() what you can't pass into apply() :)):
def apply(String, String, MetaData, NamespaceBinding, boolean, scala.collection.immutable.Seq[Node]): Elem
def unapplySeq(Node): Option[(String, String, MetaData, NamespaceBinding, scala.collection.immutable.Seq[Node])]
def child(): scala.collection.immutable.Seq[Node]
```

dubinsky:
Pull request that adds all the missing type annotations using Scala-version-dependent return types is up: https://github.com/scala/scala-xml/pull/675.
Closing this one. Thank you!

PR #657, opened 2023-03-28
Add missing type annotations that do not break binary compatibility

dubinsky:
Some missing type annotations can be added without breaking binary compatibility; they are added in this pull request, letting #655 concentrate on adding type annotations that _do_ break binary compatibility (for Scala 3).

Removed MiMa exclusions that are no longer needed.

Removed Gitter link to a non-existing room.

Added links to the compiler implementation of the `SymbolicXMLBuilder` for different versions of Scala.


SethTisue (approved):
nice

dubinsky:
> nice

thanks!

PR #658, opened 2023-04-02
Demonstrate XInclude work - and fail

dubinsky:
This pull request does not affect production code at all; it:
- adds a test dependency on `Xerces`;
- adjust tests that assume that JDK-built-in `Xerces` is the only `Xerces` on the classpath;
- adds tests demonstrating that `XInclude` works;
- adds tests demonstrating when `XInclude` fails miserably because of the old and never fixed `Xerces` bug...


SethTisue (approved):
I don't have any expertise on the specifics here, but I think you should go ahead and merge it.

dubinsky:
> I don't have any expertise on the specifics here, but I think you should go ahead and merge it.

Thank you!
Maybe you have some ideas on how we can encourage the Xerces team to fix a 20-years old bug? ;)

PR #663, opened 2023-05-22
XMLLoader returns Document

dubinsky:
- moved private def parse() from `XMLLoader` to `FactoryAdapter`;
- `FactoryAdapter.parse` returns `Document`;
- added yet another flavor of the `XMLLoader.load*` methods returning `Document`: `load*Document`;
- expressed all the other `load` methods in terms of the `load*Document`;


dubinsky:
@SethTisue @ashawley a small step towards parity between `XMLLoader` and `MarkupParser`...

PR #668, opened 2023-06-08
Capture more of the Document

dubinsky:
- handle Locator2, DTDHandler, DeclHandler, and LexicalHandler DTD and entity methods;
- introduce DtdBuilder and delegate DTD methods to it;
- introduce ElementContentModel and delegate parsing of the element content model in the DTD element declarations to it;
- XMLLoader.loadDocument() returns baseURI, XML version, encoding and standAlone flag, and DTD;
- tests demonstrating the new returns.

Now that `XMLoader` returns `Document`, there is a place to return more information abut the parsed XML document - and this change does that, bringing `XMLLoader` to parity with the `ConstructingParser` ; returned now are:
- document's `baseURI` (which `ConstructingParser` never returned and indeed couldn't return);
- XML version of the document (which `ConstructingParser` returns correctly only if it is "1.0");
- `standalone` status of the document;
- document's encoding (`ConstructingParser` seems unable to parse documents encoded in `UTF-16`);
- document's `DTD` (`ConstructingParser`  does **not** return element and attribute declarations, notations and some entities).

To preserve binary compatibility, parsed DTD is converted (with some loss of fidelity) into pre-existing `ContentModel` classes, instead of replacing them with the newly-introduced `ElementContentModel`.
To avoid capturing by the binary compatibility rules, newly introduced classes are package-private.


SethTisue:
no objection here — feel free to merge if no other reviewers appear

dubinsky:
> feel free to merge if no other reviewers appear

Thank you!

PR #672, opened 2023-06-15
Change infix method calls `o m a` to `o.m(a)`.

dubinsky:
With infix method invocations falling out of fashion - and eventually, out of validity - maybe it is time to move away from them?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -93,7 +93,7 @@ lazy val xml = crossProject(JSPlatform, JVMPlatform, NativePlatform)
           file(jarPath)
             -> url("http://docs.oracle.com/javase/8/docs/api")
         )
-      } getOrElse {
+      }.getOrElse {
         // If everything fails, jam in Java 11 modules.
         Map(
           file("/modules/java.base")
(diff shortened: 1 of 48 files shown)


Ichoran:
Even `eq`?  `foo.eq(null)` flows a lot less well than `foo eq null` to my eye.  (Personally, I find the `StringBuilder` append a lot clearer infix too, but maybe I'm weird that way.)

dubinsky:
@Ichoran 

Sorry, merged to eagerly :)

> (Personally, I find the `StringBuilder` append a lot clearer infix too, but maybe I'm weird that way.)

You are not weird, I agree that `sb append("<!ELEMENT ") append(name) append(' ')` looks cleaner than `sb.append("<!ELEMENT ").append(name).append(' ')`; but to me `s"<!ELEMENT $name "` looks cleaner still. Maybe https://github.com/scala/scala-xml/pull/673 alleviates some of the damage to readability...

> Even `eq`? `foo.eq(null)` flows a lot less well than `foo eq null` to my eye.

While I prefer `a.equals(b)` to `a equals b`, I agree that `a eq b` looks better than `a.eq(b)`... The goal of this pull request was not to codify my aesthetics though - it was to eliminate _any and all_ infix method invocations, since they are [no longer recommended](https://docs.scala-lang.org/style/method-invocation.html) and will eventually become illegal (although it'll be years before that affects us here).

I did not look into changing the code to use different methods on case-by-case basis as part of this pull request, but doing this should alleviate some of the readability issues; for instance, 9 out of 10 occurrences of `a.ne(b)` in the code are actually of the form `a.ne(null)` and could be replaced with the very readable `a != null`, couldn't they? Needless to say, your contributions in this or any other area are more than welcome :)

PR #676, opened 2023-06-20
mismatched quotes should not cause an infinite loop

### LINKED ISSUE #656: Infinite loop caused by mismatched quotes
Original comment from #306, @mbeckerle wrote:

This in XML is also causing an infinite loop:

     myAttribute=`value"

Note that the quotes are mis-matched.

dubinsky:
Resolves #656

cc: @mike-mcgann @mbeckerle


mike-mcgann:
Looks good to me, thanks!

PR #718, opened 2024-04-15
set version policy for 2.2.1 release

SethTisue:


SethTisue:
xmlJS is complaining:

```
[error] (xmlJS / versionPolicyReportDependencyIssues) Dependencies of module scala-xml:2.2.0+57-b441dbbd-SNAPSHOT break the intended compatibility guarantees 'Compatibility.BinaryAndSourceCompatible' (see messages above). You have to relax your compatibility intention by changing the value of versionPolicyIntention.
```

I guess we can call it 2.3.0 🤷

PR #742, opened 2024-11-11
Call `reset` on reusable `SAXParser` instance

lrytz:
Discussion: https://github.com/scala/scala-xml/pull/176#issuecomment-2435408637


shared/src/main/scala/scala/xml/factory/XMLLoader.scala
```diff
@@ -45,7 +45,11 @@ trait XMLLoader[T <: Node] {
   }
 
   /* Override this to use a different SAXParser. */
-  def parser: SAXParser = parserInstance.get
+  def parser: SAXParser = {
+    val p = parserInstance.get
+    p.reset()
```
dubinsky, line 50: Auto-reset on get leads to surprising changes in behavior, e.g. non-persistence of the settings on `XMLLoader.reader`, as witnessed by the changes to the test above.
Better place for auto-reset seems to be `FactoryAdapter.loadDocument()`, but it will then apply to all readers, not just the default one...
Better still would be - in my opinion - not to reuse parsers, readers and factory adapters and not to use thread-locals.

jvm/src/test/scala/scala/xml/XMLTest.scala
```diff
@@ -667,13 +667,14 @@ class XMLTestJVM {
...
-    XML.reader.setErrorHandler(new org.xml.sax.ErrorHandler {
+    val reader = XML.reader
+    reader.setErrorHandler(new org.xml.sax.ErrorHandler {
       override def warning(e: SAXParseException): Unit = gotAnError = true
       override def error(e: SAXParseException): Unit = gotAnError = true
       override def fatalError(e: SAXParseException): Unit = gotAnError = true
     })
     try {
-      XML.loadString("<a>")
+      XML.adapter.loadDocument(Source.fromString("<a>"), reader)
```
dubinsky, line 677: I'd love to make `XMLLoader.adapter`, `NoBindingFactoryAdapter` and `FactoryAdapter` private at some point (and remove some of them altogether), and add `XMLLoader.loadDocument(inputSource: InputSource, reader: XMLReader)`...

dubinsky (commented):
It is not clear to me what is gained by this change...

lrytz:
> It is not clear to me what is gained by this change...

It's really more of a guess... In a non-public codebase we have an `UnsafeXML` object that's similar to `XML` but before the safe defaults (https://github.com/scala/scala-xml/pull/177). We recently applied the same ThreadLocal caching to `UnsafeXML` as was done in https://github.com/scala/scala-xml/pull/176 and saw some SaxParserException. But I haven't followed through to understand what's going on exactly or if `reset` actually fixes it, I'll do that.

Thanks for taking a look.

shared/src/main/scala/scala/xml/factory/XMLLoader.scala
lrytz: So revert https://github.com/scala/scala-xml/pull/176?

dubinsky: I am not going to campaign for reverting it,  but I am not a fan ;)

lrytz:
> I haven't followed through to understand what's going on exactly or if `reset` actually fixes it, I'll do that.

Didn't manage to reproduce the failure we saw and suspect would be fixed by this PR.

So that leaves as justification that "it seems right" to call reset, according to the Javadocs.

SethTisue:
@dubinsky where do we stand here, any idea?

dubinsky:
> @dubinsky where do we stand here, any idea?

As mentioned in https://github.com/scala/scala-xml/pull/742#discussion_r1837437245 and https://github.com/scala/scala-xml/pull/742#discussion_r1837439563, I do not see this change as a move in the right direction (see https://github.com/scala/scala-xml/pull/765 for context).

According to https://github.com/scala/scala-xml/pull/742#issuecomment-2488781971, suspicion that this change fixes a problem has been falsified, and the only motivation for merging it is that it "feels right". 

Well, it does not feel right to me, since it leads to surprising (to me ;)) changes in behavior, e.g. non-persistence of the settings on `XMLLoader.reader`, But if this still feels right to @lrytz, let's just do it ;)

@lrytz?

dubinsky (approved):

SethTisue:
@lrytz maybe you could improve the PR description, since we'll be linking to it from the release notes?

Issue #650, opened 2023-02-27
Unused value warnings for XML literals starting with Scala 2.13.9


BennyMcBenBen:
If I am using XML literals, then I am seeing compile warnings starting with Scala version 2.13.9. I am using the sbt-tpolecat plugin so these are actually compile errors for me.

Example:

```scala
import scala.xml.Elem

case class ToDoItem(name: String) {
  def toXml: Elem = {
    <to-do-item>
      <name>{name}</name>
    </to-do-item>
  }
}
```

Compile errors:
```
[error] ToDoItem.scala:5:5: unused value of type scala.xml.NodeBuffer (add `: Unit` to discard silently)
[error]     <to-do-item>
[error]     ^
[error] /ToDoItem.scala:6:7: unused value of type scala.xml.NodeBuffer (add `: Unit` to discard silently)
[error]       <name>{name}</name>
```

Is this a bug or are the new unused warnings incompatible with XML literals? What do you recommend? Thank you.

lrytz:
The issue here is in the compiler (code it generates for XML literals), not in the xml library. Existing ticket: https://github.com/scala/bug/issues/12658

som-snytt:
Reopening in order to close it.

PR #744, opened 2024-11-25
NodeBuffer append is this.type

som-snytt:
Fixes https://github.com/scala/scala-xml/issues/650

The signature should be like https://github.com/scala/scala/blob/v2.13.15/src/library/scala/collection/mutable/StringBuilder.scala#L142 so that linting knows it does not return a new value.

The generated code has the form
```
    private[this] val xml: scala.xml.Elem = new scala.xml.Elem(null, "xml", scala.xml.Null, scala.xml.TopScope, false, ({
      val $buf: scala.xml.NodeBuffer = new scala.xml.NodeBuffer();
      $buf.&+(new scala.xml.Elem(null, "elem", scala.xml.Null, scala.xml.TopScope, false));
      $buf
    }: _*));
```

### DIFF
--- a/shared/src/main/scala/scala/xml/NodeBuffer.scala
+++ b/shared/src/main/scala/scala/xml/NodeBuffer.scala
@@ -36,7 +36,7 @@ class NodeBuffer extends scala.collection.mutable.ArrayBuffer[Node] with ScalaVe
    * @param o converts to an xml node and adds to this node buffer
    * @return  this nodebuffer
    */
-  def &+(o: Any): NodeBuffer = {
+  def &+(o: Any): this.type = {
     o match {
       case null | _: Unit | Text("") => // ignore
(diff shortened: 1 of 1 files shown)


dubinsky:
I do not understand why this change does not trigger the compatibility checker like other attempts to make the types more precise do, but I am all for this change!
Let's just wait a couple of days before merging it to give the people an opportunity to speak up.

SethTisue:
MiMa primarily analyzes bytecode. (There are a few situations in which it looks at the Scala signature, but that's not in play here.) `this.type` erases to the same thing as the old type, so MiMa (rightly) doesn't see anything to object to.

SethTisue (approved):

som-snytt:
I sympathize with the preceding question and answer. Compiler fictions FTW!

Issue #764, opened 2025-06-06
Use `FactoryAdapter` as `ContentHandler` crashes


lrytz:
This used to work a while ago

```scala
  @UnitTest
  def contentHandler(): Unit = {
    val p = defaultParserFactory.newSAXParser()
    val handler = new parsing.NoBindingFactoryAdapter
    val input = new ByteArrayInputStream("<hello>world</hello>".getBytes("UTF-8"))
    p.parse(input, handler)
    val n: Node = handler.rootElem
  }
```

This bypasses `FactoryAdapter.loadDocument`, which is used internally within scala.xml.

Currently:
  - After https://github.com/scala/scala-xml/pull/663, `rootElem` is set to `null`
  - After https://github.com/scala/scala-xml/pull/668, `endDocument` crashes with `None.get` in `xmlReader.get.getProperty(...)`

I came across some code at a customer that uses `NoBindingFactoryAdapter` as a `ContentHandler` in their internal XML library to convert to scala.xml.

I can work around it by overriding `def endDocument()`, getting out the `rootElem` and avoid calling `super.endDocument()`.

I don't know if the use case should be supported?

PR #765, opened 2025-06-06
Allow using FactoryAdapter as ContentHandler

lrytz:
Fixes https://github.com/scala/scala-xml/issues/764


lrytz:
Thanks for the pointer! In the case of jdom2, I can do

```
val source = new transform.JDOMSource(document)
XML.withXMLReader(source.getXMLReader).loadDocument(source.getInputSource).docElem
```

dubinsky:
> Thanks for the pointer! In the case of jdom2, I can do
> 
> ```
> val source = new transform.JDOMSource(document)
> XML.withXMLReader(source.getXMLReader).loadDocument(source.getInputSource).docElem
> ```

Glad it worked out!
Any suggestions for making Scala XML documentation clearer in this area are welcome ;)

PR #833, opened 2026-09-03
Enable -Yfuture-lazy-vals on Scala 3.3

### LINKED ISSUE #828: Support Java 25 without warnings
We have an application that uses various Scala dependencies, including scala-xml. When we run with Java 25 we get the following warning:
```
WARNING: A terminally deprecated method in sun.misc.Unsafe has been called
WARNING: sun.misc.Unsafe::objectFieldOffset has been called by scala.runtime.LazyVals$
```
I believe this is causes by the dependencies being compiled with a version of Scala before 3.8 and not using the flag `-Yfuture-lazy-vals`.

Scala-xml seems to use Scala 3.3.8 (so not 3.8+) and does not use `-Yfuture-lazy-vals`, and also uses `lazy val`, so using this dependency will also cause the above warnings.

Can there be a version of scala-xml released using Scala 3.8+ and/or using `-Yfuture-lazy-vals`, so that this warning can be fixed?

See https://github.com/scala/scala-parser-combinators/issues/648 for a the same issue for scala-parser-combinators.

lrytz:
Build on Scala 3.3 with `-Yfuture-lazy-vals`.

Bump the JDK for Scala 3 releases to 17.

Fixes https://github.com/scala/scala-xml/issues/828


tgodzik (approved):
LGTM!

